PR #12, opened 2019-10-11
Fix JDK naming and cleanup in ant build

sampottinger:
In response to processing#6, ensure removal of temporary archive expansions of JFX and JDKs, fix .gitignore for those downloads, and switch to the new JDK naming scheme.


build/.gitignore
```diff
@@ -1,6 +1,12 @@
 work
 javadoc
 
-macosx/jdk-*
-macosx/javafx-sdk-11.0.2
-macosx/jfx-11.0.2.zip
+macosx/javafx-sdk-*
```
sampottinger, line 4: Up to you if you want these fully spelled out or collapsed with the leading wildcard. Let me know!

benfry: Wildcard is fine; `jfx` and `javafx-sdk` are different animals?

sampottinger: The name of the folder in the JFX distribution zip is `javafx-sdk`.

sampottinger: I can change the name of the downloaded file but was trying to keep format the same between jdk and jfx. Let me know if you want me to switch!

PR #17, opened 2019-11-04
Change editor header field visibility

codeanticode:
It would be convenient if some fields in EditorHeader have protected visibility to they can be accessed from subclasses in a derived mode. This could allow customizing things like the editor tab popup menu, for example, the shader mode project introduces a "new glsl tab" to the menu:

image


benfry:
Hmm, we'd like to keep that as locked down as possible, so that folks aren't just adding weird options.

For instance, to create a GLSL tab, all you need to do is include .glsl in the file name and it should work fine.

codeanticode:
Yep, one could do just that, however is not the most intuitive alternative from the point of UI, I think, in particular for a specialized mode that may introduce support for a a specific source code language.

I wonder if a reasonable compromise between giving some degree of flexibility in the contributed modes to support other file types abut not too much room to add weird stuff affecting the consistency for the PDE's UI could be to have some public method in the EditorHeader class specifically to add a "Create X Tab"  where X corresponds to the file formats supported by the mode. Not a huge issue in any case, I can look closer into the implementation to see if this is possible.

benfry:
Things need to pass a really high bar to be added to the UI; creating GLSL tabs just isn't one of them. It's confusing for the primary audience for the PDE (“what's GLSL?”) and it's not necessary from an implementation standpoint—more code is being added to support something that already works.

The only way we've been able to keep things simple is to keep this sort of thing out.

codeanticode:
Make sense. I think that an option such as "Create X Tab" should not cause much confusion for the users if it is in a contributed mode designed to work specifically with X language. But I also see how these options would affect the simplicity of the header menu. Tabs for other formats can be created by entering the file extension anyways, so it is not an essential functionality.

Issue #16, opened 2019-10-27
static imports partly broken


clankill3r:
In 3.5.3 you could use `import static`. 
Here I use it for `Ani.init()` in setup.

```
import de.looksgood.ani.*;
import static de.looksgood.ani.Ani.*;

void setup() {
  init(this);
}
```
I use Ani as an example cause i didn't want to share a whole project to demonstrate.

With processing4 I get:

> No library found for de.looksgood.ani.Ani
> The function init(sketch_191027a) does not exist.

The 'No library found' error is not new, but not being able to use the static methods is new.

!!! I do use the latest release that can be downloaded from Sam Pottinger's site, which is dated September. I'm unable to build it myself.

sampottinger:
Hummm.... Thanks for the issue @clankill3r. I will try to take a look!

sampottinger:
Ah I see what you mean @clankill3r... This has been like half broken (half working?) for a while in mainline processing (https://github.com/processing/processing/issues/5577). :-/ I can think I can fix this! However, I am worried about Ben mentioning it was not supported (I can't find anything in the [official docs](https://github.com/processing/processing/wiki/Supported-Platforms) saying it's not supported but he [left a comment in a bug that it wasn't](https://github.com/processing/processing/issues/5577#issuecomment-406283900)). With the new ANTLR stuff, I don't see any reason on my end not to support it so I'll put a PR up but I'll leave it up to him to decide if it should be supported or if a (better) error should be thrown on encountering static imports. It does mean that import statements need to be tracked as more than just strings internally within the preprocessor and its consumers because some places can't handle anything but the fully qualified name being retained but I don't really see a huge downside to migrating that code. Logic for imports is already isolated within the build pipeline and it seems like it would be safer long term to clarify wildcards versus library name versus fully qualified name. It was confusing for me when I was reading through earlier this year when doing my own edits.

... Either way, there should either be a better error message or support for static imports. It's confusing the way it is.

sampottinger:
Hey there @clankill3r! https://github.com/sampottinger/processing4/pull/19 would resolve your issue. I'll redirect that PR to this branch shortly after a little more testing.

sampottinger:
Sorry one more update @clankill3r. I put in https://github.com/processing/processing4/pull/18. I'll post a new build up on my website in a bit depending on how that PR goes. Thanks for your patience!

clankill3r:
great! Looking forward to the new build.

[...]

PR #18, opened 2019-11-05
Add support for static imports through pdex's ImportStatement

### LINKED ISSUE #5577: import static  may throw a invalid error  "No library found for  library.class"
Modify processing so that the Error "No library found for  library.class" is not displayed when it is actually found and used by the importing program

## Expected Behavior

import static library.class;  

Static import is a feature introduced in the Java programming language that allows members (fields and methods) defined in a class as public static to be used in Java code; without specifying the class in which the field is defined.  from  en.wikipedia.org/wiki/Static_imports  They said it better than I could

No errors should be displayed.

## Current Behavior

 import static library.class;  

Works library is available , the static values are available and do not need to be prefixed 

An error is displayed in Processing IDE       "No library found for  common.Palette" 
this error is in error .  The library is being used.

removing the static option gets rid of the error but then the static values need to be prefixed with the class name.

## Steps to Reproduce
files and more words are available from  https://github.com/malcolm-gc/template_for_jar-
1.  Install the Project common_jar
2.  build the library by running CreateJarFile.bat   this will build a library common.jar
3.  Install the Project PaletteJarSketch
4.  Start PaletteJarSketch in Processing
5.   run the project 
6. Observe that an error is displalyed "No library found for  common.Palette" even though the library is being used

## Your Environment
* Processing version:  Processing 3.3.7

see  https://discourse.processing.org/t/including-shared-pde-files-across-projects/1701/47
a response by one of the users indicates that a jar created in Eclipse will not throw this error 

import static 

https://github.com/processing/processing/wiki/Supported-Platforms#java-versions  ( info provided by benfry) 

I recommend that the bug report be closed as unsupported or invalid

sampottinger:
**Summary**. 
Discontinued use of string FQN for import statement tracking and migrated to pdex's ImportStatement class which can keep track of if the import is static or not. This allows static imports to be handled correctly in PDEX, JavaBuild library resoluation, and actual Java compilation step. Resolves processing/processing#5577 and processing#16 but only in the Processing4 branch.

**Rationale**  
See https://github.com/processing/processing4/issues/16 for conversation:

> Ah I see what you mean clankill3r... This has been like half broken (half working?) for a while in mainline processing (processing/processing#5577). :-/ I can think I can fix this! However, I am worried about Ben mentioning it was not supported (I can't find anything in the official docs saying it's not supported but he left a comment in a bug that it wasn't). With the new ANTLR stuff, I don't see any reason on my end not to support it so I'll put a PR up but I'll leave it up to him to decide if it should be supported or if a (better) error should be thrown on encountering static imports. It does mean that import statements need to be tracked as more than just strings internally within the preprocessor and its consumers because some places can't handle anything but the fully qualified name being retained but I don't really see a huge downside to migrating that code. Logic for imports is already isolated within the build pipeline and it seems like it would be safer long term to clarify wildcards versus library name versus fully qualified name. It was confusing for me when I was reading through earlier this year when doing my own edits.

**Discussion**  
@benfry up to you if you want to support this. Seems doable if we switch to the (already existing) ImportStatement instead of just passing string FQNs around. If you would prefer not to support, I'll change the preprocessor to emit an easier to understand error message.

**Notes**
Supersedes https://github.com/sampottinger/processing4/pull/19.


benfry:
I'm somewhat ambivalent about including it. The previous comment about it not being supported was because it was something that 1) wasn't part of the grammar, and 2) we've never claimed to support all of Java.

I'm fine to incorporate if it doesn't make things too messy internally. What do you think?

sampottinger:
I'm modestly in favor of including it. I don't see much downside in supporting it (even if the audience that is likely to use it seems like it might be fairly small). In terms of internal code quality, it also conveniently forces clean up with regards to some ambiguity as to how imports are represented internally (some places previously used `ImportStatement` and some use a plain string but that's all unified now in order to keep track of if the import is static and / or wildcard).

benfry:
Ok, sounds good.

PR #20, opened 2019-11-07
Fix console rounding.

sampottinger:
Responding to processing#19,  there appears to be an issue with calculating the bounds of text inside  JTextPane within EditorConsole for specific font family and size  combinations. This may be related to https://bugs.openjdk.java.net/browse/JDK-8158370 or https://bugs.openjdk.java.net/browse/JDK-8156217 but resolves when changing Editor.LEFT_GUTTER from 44 to basically anything else.


benfry:
Ok, thanks.

PR #33, opened 2019-11-25
Fixed Table's use of deprecated isAccessible

### LINKED ISSUE #3: Update isAccessible() in Table to use JDK 11 reflection methods
Currently a couple warnings inside `Table` about using `canAccess()` instead of `isAccessible()`. Just need to track down the correct way to re-implement.

sampottinger:
Fixed the `Table`'s use of the deprecated `isAccessible`, finishing the `parseInto` method in the process in order to test that the new implementation is "correct". Note that there is not a published API for this method on processing.org so it is unclear if the interface's intended behavior was maintained. See https://processing.github.io/processing-javadocs/core/processing/data/Table.html#parseInto-java.lang.Object-java.lang.String-.

Resolves #3.


sampottinger:
Tested manually with a sketch and added unit test.

Issue #28, opened 2019-11-22
Cannot use rotateZ() in P3D


knupel:
These bug is still alive 
https://github.com/processing/processing/issues/5770
maybe this pull can be used, say yesss it's my first pull :)
https://github.com/processing/processing/pull/5821

sampottinger:
Thanks @StanLepunK! Thank you for your PR and I think it looks good. Would you like to open a PR against this repo or would you like me to open a PR for you? I can't do PR reviews on the mainline processing repo but I can do PR reviews here.

knupel:
Nice, I try to do that this week, because at this moment i'm like Ben... no time for me :(

sampottinger:
Thanks @StanLepunK!

benfry:
Fix has been merged.

PR #41, opened 2019-12-01
Fix RotateZ bug

knupel:
answer to this bug https://github.com/processing/processing4/issues/28


benfry:
Thanks!

knupel:
Pretty sure is the smaller pull request of the Processing history... I'm very proud of it :)

PR #42, opened 2019-12-04
Provide build and test instructions in README

### LINKED ISSUE #26: There are no build instructions
I wanted to gave it another attempt at building cause I need to set some JVM options.
However, the build instructions are gone.
After some effort I did found them here:

https://gist.github.com/sampottinger/939441d7277d318189ec81844e3ca516

But It's a bit hidden.

sampottinger:
Provide build and test instructions within the project README. Resolves #26.


benfry:
Let's put the build portion as a new README.md under `build`, though—otherwise the main page gets huge.

sampottinger:
Thanks for the feedback @benfry. I addressed in https://github.com/processing/processing4/pull/42/commits/6ed36f7abb63555e2491e98681947260505baeeb. Let me know if that looks good.

benfry:
Looks good, let's also add a link to the build instructions from that front page. I totally agree with your original sentiment of including them up front so they can be found.

PR #43, opened 2019-12-04
[WIP] Started migration to getModifiersEx.

### LINKED ISSUE #4: Switch to getModifiersEx() in PSurfaceAWT
We've not used `getModifiersEx()` in the past, because it (traditionally) has caused problems with some modifiers not being set properly. But now that it's deprecated, we need to make the switch and do a bunch of testing to make sure it's behaving the same way. 

The fix is super easy. The debugging/testing is a bit time-consuming.

sampottinger:
InputEvent.getModifiers has been [deprecated in Java 9](https://docs.oracle.com/javase/9/docs/api/java/awt/event/InputEvent.html). Migrated fields and will start testing. This is part of https://github.com/processing/processing4/issues/4.

Resolves #4.

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -1765,7 +1765,7 @@ public class Base {
           String path = e.getActionCommand();
           if (new File(path).exists()) {
             boolean replace = replaceExisting;
-            if ((e.getModifiers() & ActionEvent.SHIFT_MASK) != 0) {
+            if ((e.getModifiers() & ActionEvent.SHIFT_DOWN_MASK) != 0) {
               replace = !replace;
             }
 //            if (replace) {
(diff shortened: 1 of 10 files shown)


sampottinger:
Sorry I briefly polluted `master`. The issue on `master` has been resolved.

benfry:
Thanks, and as noted in the original issue: I tried making this switch in the past, and it caused a lot of regressions, so I had to revert to `getModifiers()` again. I don't recall all the details, but it's something for us to keep an eye on.

sampottinger:
@benfry did you just complete this?

benfry:
Not entirely… I did that changes inside core, but haven't flipped the switch inside app. The problem is that some subclasses may expect old-style modifiers in `actionPerformed()` (i.e. for `EditorButton`), and I haven't had a chance to track all that down. 

But I swapped things inside core, so that we can start debugging any regressions that come along with the switch.

PR #51, opened 2020-01-19
remove redundant boxing and casting

Minimaximize:
Removed array instantiation where parameters could be accepted as VArgs
callbackClass.getMethod(callbackMethod, new Class[] { File.class }); =>
callbackClass.getMethod(callbackMethod, File.class);

Remove redundant unboxing where appropriate
(scale instanceof Integer && ((Integer)scale).intValue() == 2) =>
(scale instanceof Integer && (Integer) scale == 2)
In this case scale will automaticaly be unboxed to it's int value

Replace protected void handleMethods(String methodName) and
protected void handleMethods(String methodName, Object[] args)
with
protected void handleMethods(String methodName, Object...args)
as it handles both cases

Remove redundant extends Object clause from type wildcards
List<Disposable<? extends Object>> reachableWeakReferences =>
List<Disposable<?>> reachableWeakReferences

### DIFF
--- a/core/src/processing/opengl/PJOGL.java
+++ b/core/src/processing/opengl/PJOGL.java
@@ -1469,7 +1469,7 @@ public class PJOGL extends PGL {
 
   @Override
   public void shaderSource(int shader, String source) {
-    gl2.glShaderSource(shader, 1, new String[] { source }, (int[]) null, 0);
+    gl2.glShaderSource(shader, 1, new String[] { source }, null, 0);
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


core/src/processing/opengl/PJOGL.java
```diff
@@ -1469,7 +1469,7 @@ public int createShader(int type) {
 
   @Override
   public void shaderSource(int shader, String source) {
-    gl2.glShaderSource(shader, 1, new String[] { source }, (int[]) null, 0);
+    gl2.glShaderSource(shader, 1, new String[] { source }, null, 0);
```
benfry, line 1472: Have you tested this? This would have been written the original way due to a compiler error.

Minimaximize: Yeah, I was able to compile without issue, and in all of my tests, the results were the same :+1: 

Feels kinda like it shouldn't be allowed, but while the individual elements of a primitive array aren't nullable, arrays are. So null is a perfectly legal value here.

PR #64, opened 2020-01-20
[WIP] Fix broken build on deprecated methods

sampottinger:
It looks like some of the deprecations broken the build as caught by Travis (I can confirm on my local). This is resolving those issues as well as a regression noticed in preproc for pdf mode.

Resolves #61

Resolves #62

Resolves #66

PR #68, opened 2020-01-22
Resolve PDF renderer parse issue in preproc.

### LINKED ISSUE #66: size call with PDF fails
Whoops! Looks like there was a regression on:

```
size(400, 400, PDF, "filename.pdf");
```

The transfer over of the 4th parameter was lost.

sampottinger:
Resolve moving the PDF renderer information provided in `size()` into settings as part of preproc. Resolves #66. Note that this supersedes the fix in #64.


sampottinger:
Added unit tests for the regression to prevent in the future. Sorry about that!

benfry:
Great; thanks!

PR #74, opened 2020-03-08
Fix uninitalized debugger in JavaEditor.

### LINKED ISSUE #73: Fail to start due to debugger
I'm trying to see what's up but it looks like `master` is broken (tested on Mac OS X). NullPointerException at `JavaEditor.java:1389`.

sampottinger:
Looks like constructor in `Editor` called from from `JavaEditor` (via super) is indirectly calling an overidden method back in `JavaEditor` (`buildModeMenu` calls `debugger().buildMenu()`) that expects debugger to have been initialized. Debugger cannot be initialized by `JavaEditor` in this case because the parent constructor (super call into `Editor`) must be at the start of JavaEditor's constructor and will not yet have returned. This switches to loading debugger on demand (upon calling `getDebugger`) which allows for the `Editor` to use overloaded methods that expect debugger to be initialized.

@benfry - I'm not sure if this is how you want to resolve this. Happy to take this in a different direction though if you would like.

Resolves #73.


sampottinger:
Sorry quick notes...

 - If debugger is useful or necessary in editor initialization, this PR suggests making `debugger` private as subclasses cannot rely on debugger having been initialized (descendants of `JavaEditor` could have overridden methods also called from `Editor` or `JavaEditor` that improperly expect `debugger` to be available). Therefore, 18653e2d938a1c2a18e69240cd221e0754ffdffa asks subclasses to either use or overload the `getDebugger` method. This is a bit of a bummer but it might be safer in the sense that `getDebugger` was already there with public access and it resolves ambiguity around the use of debugger.
 - If `debugger` is private, there is design inconsistency between `debugger` and `preprocService` in terms of visibility. However, as there is already a public `getPreprocessingService`, this suggests that subclasses should either use or overload the `getPreprocessingService`. A potential proposal is included in 926c4587d516b01cd2c8df419e0837ebdda26ddf.
 - The UI toolbar debug button no longer worked. A potential proposal is in d126c70893d8d33754af785b37c692ca71c3d9e0.

sampottinger:
Hey there! There's a few PRs outstanding (with inter-dependencies) and the build on master has been broken on Mac for some time. So, I have an integration branch over at https://github.com/sampottinger/processing4 master. That branch includes this PR. This allows one to use Processing 4 again on a Mac along with all of the recent bug fixes. For your convenience, I have made builds available at https://www.datadrivenempathy.com/processing.

benfry:
Is there a simpler way to handle this than all the refactoring? We're talking about finding the correct placement for an initialization, right? This feels like a huge response (the ambiguity of always calling a getter) just to resolve a single point where `debugger` was `null`.

sampottinger:
Hey there @benfry! Yeah, I think this came up from the January refactors for the debugger. Like I said in the original comment at the top of the PR, I'm definitely open to another way to resolve this.

Real fast to clarify the issue: Processing on master won't run due to debugger not being initialized (issues https://github.com/processing/processing4/issues/83 and #73 ). Given the way though that `Editor` and `JavaEditor` are currently structured (specifically `Editor`'s polymorphic behavior in `buildModeMenu` called from its constructor requires the debugger but the [debugger cannot be initialized without a JavaEditor](https://github.com/processing/processing4/blob/master/java/src/processing/mode/java/debug/Debugger.java#L112)).

I chose the above because it felt less invasive than some of the alternatives. That said, some other ideas I had (and may prefer) but that would impact the public facing API for Editor and its subclasses in a deeper way so didn't feel comfortable making:

 - [Have the menu built _after_ constructing the editor](https://github.com/processing/processing4/blob/master/app/src/processing/app/ui/Editor.java#L212) so that mode-specific stuff that changes the menu but that would involve changing the behavioral contract for`Editor` (likely a new method that subclasses of Editor would be expected to implement)
 - Have the debugger built explicitly using a new method that Editor subclasses would be expected to implement that could be called prior to creating the menu bar (this from [JavaEditor](https://github.com/processing/processing4/blob/master/java/src/processing/mode/java/JavaEditor.java#L107) to be called from constructor before [building the menu](https://github.com/processing/processing4/blob/2360429dc665cb53723227e0c2bc1e94ab2819b0/java/src/processing/mode/java/debug/Debugger.java#L61)).
 - [Asking Debugger to only use the Editor interface instead of JavaEditor](https://github.com/processing/processing4/blob/2360429dc665cb53723227e0c2bc1e94ab2819b0/java/src/processing/mode/java/debug/Debugger.java#L61) though I'm not sure if that has downstream effects.

All that said, I kinda interrupted you ha ha! 😅 I tackled this because it was blocking [the JSSC branch](https://github.com/processing/processing4/pull/75) originally. All that said, you did the debugger refactoring in January so, given the direction you wanted to continue to take that, I'm guessing there may be another simpler solution that I'm not seeing yet. Totally open to refactoring the other pending PRs if you want to take this elsewhere.

sampottinger:
Here's the current state of inter-dependent PRs:

```
#74 (processing fails to run on Mac)
  ^ #75 (fix JSSC crash)
  ^ #79 (issue with open recent)
  ^ #81 (save as png broken)
  ^ #88 (fix Mac file chooser + dark mode)
  ^ #91 (size rewrite fix for non-standard renderers)
  ^ #93 (fix broken tests)
  ^ #76 (Illegal reflective access)
     ^ #85 (bump JOGL)
```

benfry:
Fixed with https://github.com/processing/processing4/commit/200afc4fd85ccfacd3a178174c4f3462095b8061

sampottinger:
Re-opening but accepting https://github.com/processing/processing4/commit/200afc4fd85ccfacd3a178174c4f3462095b8061 to make it easier to merge master back into on_demand_debugger so that downstream branches are easier to continue forward.

PR #88, opened 2020-04-18
Switch to JFileChooser on Mac with VAqua

### LINKED ISSUE #77: cut/copy/paste while saving a sketch on OS X goes to the editor, not the save dialog
Not that important, but this old bug is back:

https://github.com/processing/processing/issues/1035

OSX 10.12.6
Processing 4.0a1

If you save a sketch, and paste during that proces for either the filename of the sketch or the filename of a new folder then it paste the content you copied before in the sketch itself.

### LINKED ISSUE #89: Mac Dark Mode
If we have vaqua, might as well implement it.

sampottinger:
Due to the regression in FileDialog causing command c / command v to no longer work within FileDialog on Mac, having OS X use JFileChooser for the editor. This is keeping with the precedent set in processing/processing#1035. In response, switch to VAqua for a not terrible file mac-feel chooser (which also gives us dark mode!).

Resolves #77.
Resolves #89.

Blocked on #74.


sampottinger:
I put this up because the precendent from processing/processing#1035 seems clear. That said, I'm on fence about this. I feel like there's too much benefit for using FileDialog even if copy  /  paste doesn't work. That said, the fact that you are unknowingly modifying the sketch code while attempting to copy / paste into the dialog is pretty bad.

sampottinger:
Ugh. I looked into this including a series of abandoned 3rd party libs. It doesn't look like there's an easy alternative. :(

sampottinger:
Yay great news! [Vaqua](https://violetlib.org/vaqua/overview.html) might work.

sampottinger:
Yay Vaqua works! This is good to go!

sampottinger:
Sorry continuous integration on my side caught a missing jar.

sampottinger:
Should be fixed now.

sampottinger:
Hey there! There's a few PRs outstanding (with inter-dependencies) and the build on master has been broken on Mac for some time. So, I have an integration branch over at https://github.com/sampottinger/processing4 master. That branch includes this PR. This allows one to use Processing 4 again on a Mac along with all of the recent bug fixes. For your convenience, I have made builds available at https://www.datadrivenempathy.com/processing.

clankill3r:
Using VAqua will bump the minimum requirement for processing to OSX 10.10, I don't know if this is a problem, but something to be aware of.

sampottinger:
Good call @clankill3r. I am adding in a version guard. Everything is API compatible so, if Vaqua isn't available because OS X is too old, it will just revert to the regular OS X swing behavior (with the bad JFileChooser).

app/src/processing/app/Sketch.java
```diff
@@ -841,7 +841,12 @@ public boolean saveAs() throws IOException {
     // TODO rewrite this to use shared version from PApplet (But because that
     // specifies a callback function, this needs to wait until the refactoring)
     final String PROMPT = Language.text("save");
-    if (Preferences.getBoolean("chooser.files.native")) {
+
+    // https://github.com/processing/processing4/issues/77
+    boolean useNative = Preferences.getBoolean("chooser.files.native");
+    useNative = useNative && !Platform.isMacOS();
```
benfry: Nope, the point of `useNative` is that the user wants to override the subjective choice we made about the choosers (whether formerly with Quaqua on OS X, or GTK on Linux)

app/src/processing/app/platform/DefaultPlatform.java
```diff
@@ -188,4 +206,105 @@ public float getSystemZoom() {
...
+      }
+    }
+  }
+
+  /**
+   * Determine if Vaqua swing modifications for OS X should be used.
+   *
+   * @return True if the system is compatible with Vaqua and it is preferred. False otherwise.
+   */
+  private boolean shouldUseVaqua() {
```
benfry: No need for this; `Platform.isMacOS()` is sufficient. We're already Java 11+ for this repo anyway.

build/build.xml
```diff
@@ -424,6 +424,7 @@
     <include name="app/lib/jna-platform.jar" />
     <include name="app/lib/ant.jar" />
     <include name="app/lib/ant-launcher.jar" />
+    <include name="app/lib/vaqua7.jar" />
```
benfry: Is there a URL for us to `<get>` the `vaqua7.jar` file from inside `build.xml`? (Or maybe it's time to cache it with others at download.processing.org…)

app/src/processing/app/platform/DefaultPlatform.java
sampottinger: It's not the java version... it's the version of OS X

app/src/processing/app/Sketch.java
sampottinger: Thanks @benfry. I think we [default chooser.files.native to true now right](https://github.com/processing/processing4/blob/06bf5566e8b0404ce08f2b387ef87f6d407f1887/build/shared/lib/defaults.txt#L76)? Maybe we switch this to a yes, no, auto? The native experience is probably preferred for windows and linux. The native mac experience (FileChooser) has bugs and the default non-native experience on mac (regular JFileChooser) is really bad. Vaqua (via modified JFileChooser) is... kinda native? I'm not sure how we want to change the preferences to accommodate.

I guess... as for the defaults... I'm pretty certain at this point that Vaqua should be the default for OS X (not native but also not regular JFileChooser). However, we would probably want to give a way to disable Vaqua and native FileChooser to get standard JFileChooser given that both FileChooser and JFileChooser on Mac are not in a good spot but Vaqua is non-standard. However, the native experience (FileChooser) on windows and linux is definitely preferred and should be the default.

build/build.xml
sampottinger: Yes we could use https://violetlib.org/release/vaqua/7/VAqua7.jar.

app/src/processing/app/Sketch.java
sampottinger: (but the user should be able to choose)

app/src/processing/app/platform/DefaultPlatform.java
benfry: Right, but does Java 11 even run on macOS older than 10.10? I can't imagine we'd even allow earlier than 10.13 (enforced in `Info.plist`) for the PDE itself.

sampottinger: Oh sorry 😅 I see your point. Yeah I think [we currently allow 10.08](https://github.com/processing/processing4/blob/master/java/application/Info.plist.tmpl#L41) but I don't think the AdoptOpenJDK folks test anything older than 10.10. I'll just change `Info.plist.tmpl` to reflect and remove this. Thanks @benfry!

build/build.xml
sampottinger: @benfry it's now downloading.

sampottinger:
Thanks so much @benfry! Just made some updates below. Thanks!

**Min mac version**  
Resolved with Info.plist.

**Download Jar**  
I switched this out so that it downloads from vaqua's site. However, I re-encountered an issue I had forgotten about. There's extraneous output that comes from vaqua on some macs when opening the save dialog as shown:

Screen Shot 2020-04-27 at 5 43 00 PM

The message is not useful and it's not an error. I rebuilt the jar without that message but it does mean it's a custom jar. LMK how you want to proceed.

**Preferences**  
I resolved all the questions except for how to represent this in the preferences. LMK how you want to proceed. I currently have a draft proposal in here with the following defaults:

```
# Default to the native (AWT) file selector where possible
chooser.files.native = true
chooser.files.native.macosx = false # except on mac where it's broken
```

Plus the addition of `editor.allow_vaqua` which defaults to true.

app/src/processing/app/Sketch.java
sampottinger: Put in a new proposal. See below.

sampottinger:

benfry:
Is the custom build thing solved by doing checkout/patch/build as part of the build process? Can that be done with a couple lines in `build.xml` or is it more complex than that?

Opting to merge, though a little nervous about no new releases since February 2019 (i.e. no Catalina release, and this debug message still embedded.) Didn't see a link to a source repo to see if there's any development version or sign of commits.

clankill3r:
I appreciate the hard work but I think this is a case of over engineering.
While command+c command+v now works it comes at the cost of:

- having no search field
- having no recent locations
- no Favourites or Locations (so a USB stick has to be accessed by going to Mac/Volumes for example)
- no "Show items as" or "Group items by"
- thick font rendering

benfry:
This has since been reverted, and alpha 4 is not using VAqua. Please use a new issue if there's a problem.

Issue #80, opened 2020-03-28
Saving to PNG broken in anything but the default renderer


sampottinger:
It looks like the `tiff` extension is always appended even if the requested filename ends in `.png` when using `JAVA2D`. Looks related to the recent refactor on AWT related code.

sampottinger:
`PImage:save` doesn't check for png. I see that recent edits are trying to reduce library dependency. I looked into it a bit more though and I'm not sure if there are many super great alternatives to using `ImageIO`.

sampottinger:
Cross-posting from #81: Just a quick update here... It looks like even in JavaFX and JOGL, BufferedImage is still the canonical solution. There might not be a way to avoid using it.

benfry:
Ah, this is something we need to handle differently in the renderers themselves… i.e. if they're using AWT objects (i.e. `Image` or `BufferedImage`) they need to create `ImageAWT` objects. The refactor was about moving all that AWT cruft out of the main classes so that we have little or no AWT in the main `processing.core` package. 

JOGL and JavaFX also have their own more efficient ways of handling image objects, our use of AWT in those was just a crutch to get things working quickly; this would provide a pathway for others to build out that improved image support.

sampottinger:
Thanks @benfry. I'm sorry could you say more about how JavaFX / JOGL would handle this? Maybe just taking JavaFX first... looking around online, [it's not clear how to do this in JavaFX without using the AWT Image classes](http://java-buddy.blogspot.com/2013/04/save-canvas-to-png-file.html). I may be missing something obvious though.

sampottinger:
In the meantime, @benfry opened PR #97 for just AWT.

sampottinger:
Also, if you did want that for FX2D, see #98

benfry:
> I'm sorry could you say more about how JavaFX / JOGL would handle this? Maybe just taking JavaFX first... looking around online, [it's not clear how to do this in JavaFX without using the AWT Image classes](http://java-buddy.blogspot.com/2013/04/save-canvas-to-png-file.html). I may be missing something obvious though.

Nope, not missing anything, sorry for the confusion… For now it's necessary to use `PImageAWT`. But the idea is that we get the AWT-specific code out of the main classes, and then the usage of AWT is determined independently by individual renderers… i.e. for now, JavaFX and JOGL just need to create `PImageAWT` objects instead of `PImage` if they want to do saving or have access to the AWT code.

[...]

PR #97, opened 2020-04-28
Allow PGraphicsJava2D to save using ImageIO via PImageAWT.

sampottinger:
Related to #80, allow PGraphicsJava2D to save a PNG via ImageIO. Note that this does not allow png saving in other renderers.


sampottinger:
~~Travis CI found some issues here. Looking into it.~~ Never mind! Looks like we should be ok.

sampottinger:

benfry:
This was cleaned up in beta 5.

PR #98, opened 2020-04-28
Implement FX2D save through SwingFXUtils.

sampottinger:
Related to #80, implement FX2D save through SwingFXUtils. This still uses PImageAWT but it looks like BufferedImage is still the canonical solution for FX.


sampottinger:

PR #99, opened 2020-05-10
Warn users or provide auto-fix when `frame` is used in a sketch

urbanskimichal:
Work done:
- Added error message when unrecognised `frame`  occurs during compilation.
- Added translatable message about the change in editor. 
- Updated editor to show `Changes in 4.0` document.

For now I don't think I'm ready for the auto-fix, but can think about it in the future.
This is my first contribution so I'm waiting for suggestions or opinions.

### DIFF
--- a/app/src/processing/app/ui/Editor.java
+++ b/app/src/processing/app/ui/Editor.java
@@ -1310,6 +1310,12 @@ public abstract class Editor extends JFrame implements RunnerListener {
     }
   }
 
+  static public void showChangesV4() {
+    if (!Base.isCommandLine()) {
+      Platform.openURL("https://github.com/processing/processing4/wiki/Changes-in-4.0");
+    }
+  }
+
 
   // . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . .
 
--- a/java/src/processing/mode/java/CompileErrorMessageSimplifier.java
+++ b/java/src/processing/mode/java/CompileErrorMessageSimplifier.java
@@ -242,7 +242,13 @@ public class CompileErrorMessageSimplifier {
 
       case IProblem.UndefinedName:
         if (args.length > 0) {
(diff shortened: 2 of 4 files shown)


java/src/processing/mode/java/CompileErrorMessageSimplifier.java
```diff
@@ -242,7 +242,13 @@ public static String getSimplifiedErrorMessage(IProblem iprob, String badCode) {
 
       case IProblem.UndefinedName:
         if (args.length > 0) {
-          result = Language.interpolate("editor.status.undef_name", args[0]);
+          final String undefinedName = args[0];
+          if (undefinedName.equals("frame")) {
+            result = Language.interpolate("editor.status.item_removed", undefinedName, "surface");
+          }
```
sampottinger, line 248: Minor request... I think we have been doing `} else {`

app/src/processing/app/ui/Editor.java
```diff
@@ -1310,6 +1310,12 @@ static public void showChanges() {
     }
   }
 
+  static public void showChangesV4() {
```
sampottinger, line 1313: Thank you so much for this @urbanskimichal! What are your thoughts on replacing `showChanges` outright?

sampottinger (changes requested):

app/src/processing/app/ui/Editor.java
sampottinger: Actually how about doing a showChangesV3 and showChangesV4 (same for handleCrustyCode). See https://github.com/urbanskimichal/processing4/pull/2.

sampottinger:
I made the above changes ahead of an integration branch build. It's ok if we go a different direction. See https://github.com/urbanskimichal/processing4/pull/2. Thank you again @urbanskimichal!

sampottinger:

PR #100, opened 2020-05-10
Editor text copy - new behavior

urbanskimichal:
This is proposal of the feature that I personally miss the most in PDE - copying / cutting whole lines.
When no text is selected and one presses **Ctrl+c** (Cmd+c) or **Ctrl+x** (Cmd+x), the whole text in the line on which the caret is, will be coped or cut.
It allows for easy lines duplicating etc.


app/src/processing/app/syntax/JEditTextArea.java
```diff
@@ -1212,6 +1212,20 @@ public final void selectAll()
     select(0,getDocumentLength());
   }
 
+  /**
+   * Selects all text in the given line.
+   * @param line The line number to select all text in it.
+   */
+  public final void selectLine(final int line)
```
sampottinger, line 1247: Hey @urbanskimichal! I think I'm in favor of the change if it isn't confusing to new users (I don't think it would be, it's common behavior for a lot of editors). However, could you say more about why you made this public?

sampottinger (changes requested):

sampottinger:
Hey there @urbanskimichal! I am doing a integration build so I modified the above to private. It's still ok to change back but wanted to make that edit before bringing this branch into integration. That is not reflected in your fork.

sampottinger:
See https://github.com/urbanskimichal/processing4/pull/1

sampottinger:

benfry:
If we can get a rebased version with that change to `private`, I can incorporate this.

urbanskimichal:
I resolved conflicts, should I do anything else? What does it mean "to private"?

sampottinger:
I think @benfry wants selectLine to be made into a private method instead of a public method? (If it’s possible)

benfry:
All set for beta 4; thanks!

urbanskimichal:
So the change is included?

benfry:
It will be in beta 4, the next release.

PR #105, opened 2020-06-29
Insert int after color instead of before to allow visibility modifier.

### LINKED ISSUE #104: Color as return type is broken
Getting a syntax error here:

```
public void setup() {
  fill(getColor());
  noStroke();
  size(100, 100);
  ellipseMode(CENTER);
  ellipse(50, 50, 50, 50);
}

color getColor() {
  return #333333;
}

public void draw() {
}
```

sampottinger:
During ANTLR traversal, insert `int` after `color` instead of before. This allows visibility modifier (like `public`) to get inserted at the correct spot (prevents `int public`). Resolves #104.


sampottinger:

PR #112, opened 2020-07-11
[WIP] Add access to JavaFX for tools.

### LINKED ISSUE #110: Some 3.x Tools not working because JavaFX isn't on the classpath
## Description
As stated, Java 11 comes without JavaFX support, which needs to be provided separately.  I found it out by trying some tools, for example, the "Perceptual Colorpicker", which didn't work.
## Expected Behavior
The "Perceptual Colorpicker" should run without problems, like normal Processing scripts with JavaFX.

## Current Behavior
A <code>java.lang.NoClassDefFoundError</code> exception is thrown, because the class <code>javafx/scene/Parent</code> is currently missing.

## Steps to Reproduce
1. Start the Processing IDE
2. Try to open the "Perceptual Colorpicker" under Tools -> Perceptual Colorpicker

## Your Environment
* Processing version: 4.0a1
* Operating System and OS version: macOS Mojave 10.14.6
* Other information: I am using a MacBook Pro Late 2013.

JavaFX is no longer included in Java 11.

sampottinger:
Resolves #110.

### DIFF
--- a/app/build.xml
+++ b/app/build.xml
@@ -35,6 +35,14 @@
 	   encoding="UTF-8"
 	   includeAntRuntime="false"
 	   classpath="../core/library/core.jar;
+                      ../core/library/javafx-swt.jar;
+                      ../core/library/javafx.base.jar;
+                      ../core/library/javafx.controls.jar;
+                      ../core/library/javafx.fxml.jar;
+                      ../core/library/javafx.graphics.jar;
+                      ../core/library/javafx.media.jar;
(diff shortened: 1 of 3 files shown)


sampottinger:
Please don't merge yet! Please see discussion at #110.

benfry:
Thanks for this; was able to work it into a fix for the JavaFX problem that also helps us avoid duplicating `core.jar` when building Processing.app on macOS.

PR #116, opened 2020-08-01
Proposal - Extract PImage.save implementations and delegate to internal ImageWriter

Minimaximize:
Related to #80

Extract save implementations from `PImage` (and subclasses) and instead delegate to an implementation of `processing.core.io.ImageWriter`.

eg. `TiffImageWriter` for encapsulating logic for writing Tiff image files.

**Benefits**
- This removes the responsibility of image serialization away from `PImage`, reducing complexity.
- Makes `java.awt` a transient dependency, making it easier to remove in future.
- Makes it possible to write unit tests for individual file formats.


benfry:
Closing this out; the image code was streamlined a bit in beta 5 so we should be in better shape.

PR #125, opened 2020-08-20
Zoom dialog fonts based on user scale setting

### LINKED ISSUE #111: Implement font scaling for the entire editor based on processing zoom preference
The only way to resize the text in the Processing UI under Windows is to use the Processing-native scaling in the preferences screen. Here' the details from #102:

> So, you are seeing the results of the fix to #31. Re-enabling sun.java2d.uiScale.enabled and everything works well (👍) but only if your scaling is a multiple of 100% (👎) otherwise the fractional text width problem comes back as described in #21 and the editor becomes (arguably) unusable. I tested this with the latest build from adoptopenjdk and an updated version of Windows 10. Indeed, it's still around (😭). It's a little tricky too because the issue is managed in config.xml for Windows for us. Interestingly, while I can't be sure, we might not be the only ones confronting this issue (note the date). It's hard to say where the blame is for this issue (existing in the liminal space that is created in the relationship between Swing and Windows) except to say that it's related to JEP 263 and the fact that there's not a super great way to respond to the screen scaling within the application (as opposed to JVM) layer. I'm not sure why but the JNA way of finding the display scaling isn't reliable... but as far as I can tell is the only way to do it.

> So... yikes... Not great. Really, I think the question comes down to accessibility. My recommended approach would be to keep sun.java2d.uiScale.enabled disabled for the reasons discussed in #21 / #31 and then (😬) implementing custom font size based on interface scaling across all the UI so that low vision users can at least use "native" processing scaling for a good experience with support for 150% zoom under the assumption that sun.java2d.uiScale.enabled=false. This is because we can't really guarantee a solution to fractional character width calculation under any other set of configurations.

I will mention that display scaling isn't "`(Recommended)`" at time of writing per the Windows-native display settings screen (see below). Unfortunately, I think this is necessary from an accessibility standpoint (especially since Windows-native display scaling can cause other apps to misbehave so it's not really accessibility-friendly either) but... it is a substantial amount of work. We should also probably enable 150% scaling if possible now that (I think?) our fractional character width handling has gotten more sophisticated.

Screen Shot 2020-07-10 at 4 53 16 PM

sampottinger:
Updates the default fonts after applying the look and feel to allow users' scale settings to zoom on all dialogs, not just main editor window. This helps with accessibility. Resolves #111.

### DIFF
--- a/app/src/processing/app/platform/DefaultPlatform.java
+++ b/app/src/processing/app/platform/DefaultPlatform.java
@@ -26,9 +26,12 @@ package processing.app.platform;
 import java.awt.Color;
 import java.awt.Component;
 import java.awt.Desktop;
+import java.awt.Font;
 import java.awt.Graphics;
 import java.io.File;
 import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
 
 import javax.swing.Icon;
 import javax.swing.UIManager;
@@ -38,6 +41,7 @@ import com.sun.jna.Native;
 
 import processing.app.Base;
 import processing.app.Preferences;
+import processing.app.ui.Toolkit;
 
 
 /**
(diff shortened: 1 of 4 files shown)


benfry:
Thanks, in for alpha 2.

PR #132, opened 2020-09-16
Doclet and code comments updates for the new Processing website

runemadsen:
This PR includes two major changes to the Processing 4 code:

1. **Added `doclet` folder** with all the files necessary for parsing the Javadoc comments in the `.java` files into `.json` files for the website. We also added a `doclet/README.md` file with instructions for running the doclet. The Doclet code is based on existing code from the `processing-docs/java_generate` folder, which means that it uses the `javadoc` and `doclet` API's that seem to be deprecated with newer versions of Java. However, this is simply stated in a warning, and it seems unlikely that these API's will be deprecated in the foreseeable future.

2. **Updated all code comments** to reflect the structure needed for the new website. We edited the files in `core/src/processing` and `java/libraries` with the following changes:
   1. Remove auto-generated text
   2. Update all main descriptions to the reflect the content of the old XML files in the `processing-docs` repo
   3. Change `@brief` to `@webBrief`
   4. Add `@webBrief` wherever missing based on the main descriptions
   5. Added categories and removed duplicated subcategories in the `@webref` tag to prevent duplication (such as `@webRef serial:serial` was changed to `@webRef serial`

Let us know how this all looks! Once this is merged in, we can begin to build the website on top of the output from the `doclet`.


benfry:
Ok, quick questions just to make sure we're set:

1. We'll no longer be auto-generating these, correct? This will be the reference of record?
2. I see that there are many changes that replace older reference in the code with what was in the XML files (and perhaps hadn't been updated yet). That's useful (hopefully!) though there are also some methods that had additional technical information that was not part of the reference. Are those still intact?
3. How do we split the technical details that are relevant to the javadoc and need to remain in the source vs. the portions that should go into the reference?

runemadsen:
Thanks! 

> We'll no longer be auto-generating these, correct? This will be the reference of record?

Correct. This is now the reference of record and all changes should be made against the code comments.

> I see that there are many changes that replace older reference in the code with what was in the XML files (and perhaps hadn't been updated yet). That's useful (hopefully!) though there are also some methods that had additional technical information that was not part of the reference. Are those still intact?

We manually went through and merged all the XML descriptions with the code comments leaving the advanced parts intact, so everything should be good to go. I wouldn't say it's an absolute guarantee since it's a lot of functions, but we feel pretty good that we didn't delete existing information.

> How do we split the technical details that are relevant to the javadoc and need to remain in the source vs. the portions that should go into the reference?

We kept the way the existing doclet worked, which means that you need to have a `<h3>Advanced</h3>` heading, which will split it. We just pushed a small update to the PR to fix this in a few comments.

@benfry Since we didn't change a lot of things in the comments, we're thinking that the javadoc generation still works, and we did successfully export the API docs, but perhaps you want to give it a try against this PR to make sure that the javadoc site still works?

REAS:
FYI, the Javadoc build starts on line 1708 of build.xml. I think we'll need to make some changes there as well: https://github.com/processing/processing/blob/master/build/build.xml

benfry:
Great, thanks @runemadsen. And yes, @REAS is correct re: the hacky Javadoc section of build.xml likely needing an update. It's probably not been touched in years.

REAS:
Good news! I was able to build from source with these changes. I haven't looked into the `doclet` stuff yet.

REAS:
I was able to build the JSON files for the Processing and the relevant libraries. It was quick!

benfry:
Looks like the script is leaving in trailing spaces, can that be fixed for future updates? The moment that `PImage.java`, `PGraphics.java`, or `PApplet.java` are re-saved (and trailing space removed automatically by the editor), it shows several edits due to the changes.

runemadsen:
Got it. We should fix that. We'll take a look!

runemadsen:
@benfry We just took a look at this, and we're not entirely clear on the problem. Can you elaborate on what's happening? Is this when you build Processing (which we didn't touch)? When you say that trailing white space are automatically removed by the editor, what editor and what auto-format are you using? Thanks!

benfry:
So after this was merged, the new comments introduced a lot of trailing whitespace. If we're ever going to run that again, we should make sure that it's not added back. 

But maybe it's a moot point if we won't ever do a process like this again? Since the JavaDoc comments will now be used to generate everything else?

(The concern was that in the past, the JavaDoc comments were generated, so each time that process ran, it made a mess of the source until we fixed the trailing whitespace issue.)

runemadsen:
Thanks Ben! I now see what you're saying. We're not entirely sure where this came from - perhaps our code editor when we manually ported over the updated descriptions - but it's odd since [some files](https://github.com/processing/processing4/commit/48397b815aad26e2041fb5d239bb4313cf35c0d2) have the trailing whitespace even though we didn't even touch them. Just to be sure, we pulled down the updated repo, built processing, ran the doclet, and diffed the files with Git and saw no changes. Since we won't need to run through the comments manually again, I think we're good for now on this.

benfry:
Ok, thanks.

PR #134, opened 2020-09-24
Fixes color selector pt localized menu name in PDE_pt.properties

villares:
Côr -> Cor
As in https://github.com/processing/processing/pull/6097


benfry:
And thanks again!

PR #139, opened 2020-09-25
Move PdePreprocessIssueException to test.

### LINKED ISSUE #130: Move PdePreprocessIssueException to the test package
This code seems to be only used by the testing classes, so it doesn't appear to be used until the tests (or a dist) is attempted. Let's move it to the test package to avoid confusion.

sampottinger:
Resolves #130


benfry:
Thanks!

PR #147, opened 2020-10-15
Small change in Portuguese translation

Maharal:
"refrência" to "referência"


benfry:
I'll have to trust you; thanks!

PR #152, opened 2020-11-21
Use preproc to rewrite smooth and noSmooth.

### LINKED ISSUE #149: Regression: smooth(4) shows “smooth() can only be used inside settings()” error
Looks like `smooth()` not properly being hoisted out of `setup()` in 4.x, inside the PDE this results in “smooth() can only be used inside settings()” in the console.

```processing
void setup(){ 
  size(300,300, P2D);
  smooth(4);
}

void draw(){
  background(0);
  fill(255,0,0);
  ellipse(100,100,100,100);
  fill(0,255,0);
  ellipse(150,150,100,100);
}
```

sampottinger:
Fixes issue around rewriting smooth / noSmooth to settings. Resolves #149 as reported by @benfry. Added unit tests for these cases. Also cleans up error reporting on rewrites with params and duplicate logic to prevent accidental rewrite for calls to user-defined methods (calls to methods matching the name of methods whose calls must be rewritten but were defined by the user regardless).


sampottinger:
Note that most of the files changed are tests.

benfry:
Thanks @sampottinger

PR #166, opened 2021-01-22
Make parseJSONObject/Array return null when parsing fails

### LINKED ISSUE #165: parseJSONObject() and parseJSONArray() do not return null when parsing fails
## Description
Both parseJSONObject() and parseJSONArray() throw a RuntimeException if the provided JSON String is not valid.

## Expected Behavior
The documentation states: 

> If the String does not contain JSONObject data or cannot be parsed, **a null value is returned**.

Both functions should therefore return null in the event of a parsing error. This would not prevent the stackTrace from being displayed in console so that the user can understand the reason.

## Steps to Reproduce
1. Pass an invalid JSON String to either parseJSONObject() or parseJSONArray()

```java
JSONObject object = parseJSONObject("invalidJSONString"); 
println(object);
```

2. Run the sketch. A RuntimeException is thrown, stopping execution unless we use a try..catch.

## My Environment
* Processing version: 4.0a3
* Operating System and OS version: MacOS 10.14.6

## Possible Solution
In PApplet.java, we could catch the RuntimeException inside the parseJSONObject/Array(). Something like this:

```java
public JSONObject parseJSONObject(String input) {
  try {
    return new JSONObject(new StringReader(input));
  } catch (RuntimeException e) {
    e.printStackTrace();
    return null;
  }
}
```

I will open a pull request.

denezth:
Fixes #165. 

The RuntimeException (thrown when the JSON String could not be parsed) is now caught in both parseJSONObject() and parseJSONArray() so that null can be returned. I also fixed the indentation of parseJSONArray's doc.

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -5814,7 +5814,12 @@ public class PApplet implements PConstants {
    * @see PApplet#saveJSONObject(JSONObject, String)
    */
   public JSONObject parseJSONObject(String input) {
-    return new JSONObject(new StringReader(input));
+    try {
+      return new JSONObject(new StringReader(input));
+    } catch (RuntimeException e) {
+      e.printStackTrace();
+      return null;
+    }
   }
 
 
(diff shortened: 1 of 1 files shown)


benfry:
Thanks! Unfortunately the indentation change meant there was a conflict, but I've just added your change by hand. Much appreciated!

benfry:
Merged here: https://github.com/processing/processing4/commit/53246fb23d519afa9e746ff1bba96b48e10b4ad7

PR #172, opened 2021-02-14
Remove two redundant variable assignments in PShader

yblake:
My two cents contribution...


hamoid:
1 cent per line XD

benfry:
Sometimes @codeanticode likes to set a variable twice to make sure it *really* gets set.

codeanticode:
Even three times, just in case :-)

PR #175, opened 2021-02-25
Fix undo/redo when switching between tabs

nking07049925:
Switching a code tab never stops the current text edit history, resulting in undo/redo making changes to tabs other than the current one. 
There is also only one caret undo/redo stack per Editor, resulting in an incorrect caret placement when switching between tabs and undoing / redoing.
This was mentioned in [this comment](https://github.com/processing/processing/issues/4775#issuecomment-569511757) in issue processing/processing/issues/4775.

I added `endTextEditHistory();` to `Editor::setCode` and `caretUndoStack` and `caretRedoStack` to `SketchCode` so every tab has a unique set of caret positions. Editor's caret undo/redo stacks are changed in `setCode` according to current code tab the same way it currently works for `Editor.undo`.

This is my first time making a Pull Request to a public repository so any assistance would be appreciated.

### DIFF
--- a/app/src/processing/app/SketchCode.java
+++ b/app/src/processing/app/SketchCode.java
@@ -25,6 +25,7 @@
 package processing.app;
 
 import java.io.*;
+import java.util.Stack;
 
 import javax.swing.text.BadLocationException;
 import javax.swing.text.Document;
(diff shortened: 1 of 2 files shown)


benfry:
Whoa… does that mean that https://github.com/processing/processing/issues/4775 is fixed? That would be amaaaazing.

Issue #181, opened 2021-03-14
Mousereleased mousebutton on default renderer


GanerCodes:
When using mouseReleased, on the default renderer (not P3D for example) mouseButton is always zero. You can still see a button type (1, 2, 3) if you pass in MouseEvent and call getNative() on it.

oseph:
Can confirm this behaviour on my end using macOS.

I'm looking at `handleMouseEvent(MouseEvent event)` method in PApplet and can't seem to make sense of why this might be happening. I'm new to contributing but I'd be happy to investigate further if someone can point me in the direction.

While I was there I also noticed code to allow macOS users to ctrl-click as right mouse, but this does not seem to work either.

behreajj:
Hi @oseph , maybe in [PSurfaceAWT](https://github.com/processing/processing4/blob/d60cc830ddca613a13e42151112d47502e8c1de9/core/src/processing/awt/PSurfaceAWT.java#L1319), due to this [update](https://github.com/processing/processing4/commit/26d04aed77fe5392a7d7868afb8f83846cafd09f#diff-d703b50193b4b119b4f02f9126610c12e6e4273c35cec675360b55166128dd86) on Jan. 20. My hack to deal with this, based on @GanerCodes suggestion, is [here](https://github.com/behreajj/CamZup/blob/15688c585e0c570eb4980a1564b8ab847935176a/src/camzup/pfriendly/ZSurfaceAwt.java#L590). I was not able to test on Mac.

`mouseClicked` is impacted as well as `mouseReleased`.

oseph:
Alrighty, I _think_ I fixed this bug in #188. Feel free to try it out. Works on my end but I only tested on macOS. Give it a spin if you're using Linux/Windows!

behreajj:
Hi @oseph,

Thanks for all your hard work. If I were you I would split this into 3 separate pull requests: one to fix the mouse released/clicked issue, one to fix the mac os control click issue and one to introduce new constants for mouse events. I'm only a random person on the Internet, mind you, trying to keep his own library working. This is just my 2c.

oseph:
@behreajj thank you! That sounds totally reasonable to me. I'm still a little new to this whole contributing thing, so would be happy to do it the proper way. I'll close/cancel this pull request and split it into 3 separate ones.

benfry:
Fixed for the release after 4.0 alpha 5 (either 4.0 alpha 6 or beta 1).

PR #188, opened 2021-03-23
Bugfix mouseReleased

oseph:
This should address #181 where `mouseReleased()` misreports all events with a `mouseButton` value of `0`

As well, `ctrl-leftClick` for macOS users should be properly recognized as a secondary(right) click.

I also updated PConstants to include `MOUSE_LEFT, MOUSE_MIDDLE, MOUSE_RIGHT` instead of the `LEFT,CENTER,RIGHT` constants. The current constants are associated with keyboard events such as `KeyEvent.VK_LEFT` and returned values that didn't make much sense, e.g. left click returning a mouseButton value of 37?

In short, mouseButton returns 1,2,3 (leftclick ,middleclick, rightclick).

As far as I can tell on my end, everything behaves properly for sketches using the default renderer, P2D, and P3D. I've only tested this on macOS (11.2.3). Would love it if anyone out there using Linux/Windows could confirm things behave on their end.

### DIFF
--- a/core/src/processing/awt/PSurfaceAWT.java
+++ b/core/src/processing/awt/PSurfaceAWT.java
@@ -1276,6 +1276,22 @@ public class PSurfaceAWT extends PSurfaceNone {
     // the 'amount' is the number of button clicks for a click event,
     // or the number of steps/clicks on the wheel for a mouse wheel event.
     int peCount = nativeEvent.getClickCount();
+    int peButton =  nativeEvent.getButton();
+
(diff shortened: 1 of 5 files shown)


benfry:
Thanks for the attempt. As someone mentioned in #181, definitely better if you can submit patches for the issue at hand without extra changes. For instance, the short constant names (`LEFT`, `RIGHT`, `CENTER`) are a feature, not a bug—you'll notice a similar level of brevity throughout the entire API. And regardless, I can't just rename them—it would break code and examples for many, many thousands of people.

PR #190, opened 2021-04-08
Fixed constructor for classes, name for fields, and made 'other' type…

Anadroid:
This PR:

1. fixes missing `)` in constructor name for classes
2. removes `[]` in field names
3. changes type `field` to type `other` for references that don't belong to classes (such as constants, functions and fields)
4. adjusts `processingrefBuild.sh` so that it overwrites individual files instead of folders


Anadroid:
This was meant for our fork. We will submit a PR when it's ready.

Issue #197, opened 2021-05-20
createFont() changes char array 


ertdfgcvb:
## Description
When passing a `char[]` as last parameter to `createFont()` the array is changed (sorted?) in place.

## Expected Behavior
The array shouldn't be touched by the function.

## Steps to Reproduce
```
void setup(){
    char[] c = {'c', 'b', 'a'};
    PFont f = createFont("Helvetica", 16, false, c);    
    println(c);
}
[0] 'a'
[1] 'b'
[2] 'c'
```
## Your Environment
* Processing version:
4.0a3
* Operating System and OS version:
* MacOS 10.15.7

benfry:
Yikes, yeah—bad practice. Thanks for identifying the problem and fixing.

Merged and updated for 4.0a4.

PR #198, opened 2021-05-26
fix for createFont bug #197

jaysonh:
now createFont does not sort the original charset array. Fix for bug #197 

New code makes a copy of the array and then sorts the copy.


benfry:
Thanks!

PR #199, opened 2021-05-28
Calls to `unregisterMethod` should null out object/method pairs instead of shortening the object and method arrays

hx2A:
This PR is to fix a bug I found in Processing related to calls to `unregisterMethod`.

The problem occurs if there are multiple registered objects and they try to unregister themselves during their method calls. I discovered this bug while working with the Processing Video library. The `Movie` class's [dispose() method](https://github.com/processing/processing-video/blob/321de067220763aedae8aa18ba3115448d381e91/src/processing/video/Movie.java#L117) unregisters the object's dispose call during its dispose call.

The `RegisteredMethods`'s `handle` method will loop through the object/method pairs, but if while doing that the `RegisteredMethods`'s `remove` method gets called, the `object` and `method` arrays will change size, and the for loop will skip an object/method pair. This is a problem for the Video library if multiple Movie objects are playing and the sketch exits. Dispose gets triggered and expensive Movie objects continue operating (and the JVM isn't getting shut down).

To reproduce the problem, try this code:

```
public class TestObject {
  
  private PApplet parent;
  private String name;

  public TestObject(PApplet parent, String name) {
    this.parent = parent;
    this.name = name;

    parent.registerMethod("post", this);
  }

  public void post() {
    System.out.println("in post for object " + name);

    if (parent.frameCount == 10) {
      parent.unregisterMethod("post", this);
    }
  }
}
```

With a sketch like this:

```
TestObject test1;
TestObject test2;

void setup() {
  test1 = TestObject(this, "object 1");
  test2 = TestObject(this, "object 2");
}

void draw() {
  if (frameCount == 20) {
    exit();
  }
}
```

The fix for this is for `remove` to null out the object/method pairs instead of shortening the arrays and for `handle` to check for nulls.

Alternatively one could leave `remove` alone and have `handle` make copies of the `method` and `object` arrays, so the arrays wouldn't change size while the for loop is iterating through the array items. But that would mean there would be 2 array copies every time `handle` gets called, which seems inefficient to me.

I'm happy to explore other fixes if this is not sufficient.

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -1367,6 +1367,10 @@ public class PApplet implements PConstants {
 
     void handle(Object[] args) {
       for (int i = 0; i < count; i++) {
+        // skip nulled out object/method pairs
+        if (methods[i] == null || objects[i] == null) {
+          continue;
+        }
         try {
           methods[i].invoke(objects[i], args);
         } catch (Exception e) {
@@ -1412,24 +1416,17 @@ public class PApplet implements PConstants {
 
 
     /**
-     * Removes first object/method pair matched (and only the first,
+     * Null out the first object/method pair matched (and only the first,
      * must be called multiple times if object is registered multiple times).
-     * Does not shrink array afterwards, silently returns if method not found.
+     * Does not shrink array. Silently returns if method not found.
      */
 //    public void remove(Object object, Method method) {
 //      int index = findIndex(object, method);
     public void remove(Object object) {
       int index = findIndex(object);
       if (index != -1) {
-        // shift remaining methods by one to preserve ordering
-        count--;
-        for (int i = index; i < count; i++) {
-          objects[i] = objects[i+1];
-          methods[i] = methods[i+1];
-        }
-        // clean things out for the gc's sake
-        objects[count] = null;
-        methods[count] = null;
+        objects[index] = null;
+        methods[index] = null;
       }
     }
 
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -1366,9 +1366,11 @@ public class PApplet implements PConstants {
 
 
     void handle(Object[] args) {
+      boolean nullsDetected = false;
       for (int i = 0; i < count; i++) {
         // skip nulled out object/method pairs
         if (methods[i] == null || objects[i] == null) {
+          nullsDetected = true;
           continue;
         }
         try {
@@ -1392,8 +1394,19 @@ public class PApplet implements PConstants {
           }
         }
       }
-    }
 
+      if (nullsDetected) {
+        int index;
+        while ((index = findIndex(null)) >= 0) {
+          // shift remaining methods by one to preserve ordering
+          count--;
+          for (int i = index; i < count; i++) {
+            objects[i] = objects[i+1];
+            methods[i] = methods[i+1];
+          }
+        }
+      }
+    }
 
     void add(Object object, Method method) {
       if (findIndex(object) == -1) {


dzaima:
Worst-case scenario with this change is that you repeatedly add and remove a handler, ending up with a large array of almost all `null`s, which gets slow quickly. The traditional thing to do would be to leave `remove` alone and just iterate through the list backwards in `handle`, though that may not be acceptable if the evaluation order matters.

hx2A:
A large array of `null`s would slow things down, but that worst case scenario would be rare, so I think this is better than two Array copies every single time `handle` is called.

> The traditional thing to do would be to leave `remove` alone and just iterate through the list backwards in `handle`, though that may not be acceptable if the evaluation order matters.

I agree!! But order does matter for these registered methods, so that doesn't work. Also, there is the edge case that an object tries to remove a different object's registered method.

But here's an idea: what if I added a flag to indicate that there are nulls in the array, and outside that loop, clean up the arrays to remove them and decrease the `count`? That would prevent the accumulation of `null`s. Let me write and push that code to make the idea more clear.

hx2A:
@dzaima Now the `handle` method checks for nulls and removes them if needed.

benfry:
Wait, so you're removing the handler in the middle of the handler being called? Basically removing an entry from the list *while* it's being iterated over. That seems like a terrible idea; is there no other place to do it? 

Arguably, the “bug” here is more about the fact that it fails silently since that chunk of code is so old that it's not using standard Collections classes (due to performance reasons), so you're not getting an error saying “don't modify this list while iterating.” 

The alternative would be to create a copy of the handlers on each iteration, just in case someone messes tries to remove a handler while it's being iterated upon. But that makes the code slower for everyone, so back to the original question—is there not a better way to approach this? (There may not be; this is just not a situation I've thought about specifically.)

benfry:
I've cleaned up the code a bit with https://github.com/processing/processing4/commit/24a56e5d3d1e1c5f0cf5c016edfb2a44bcd3d7fa but open to suggestions on removing handlers while they're being called.

hx2A:
> Wait, so you're removing the handler in the middle of the handler being called?

Yes, that's what the Processing Video library does in the [dispose() method](https://github.com/processing/processing-video/blob/321de067220763aedae8aa18ba3115448d381e91/src/processing/video/Movie.java#L138).

> The alternative would be to create a copy of the handlers on each iteration, just in case someone messes tries to remove a handler while it's being iterated upon. But that makes the code slower for everyone

Agreed, making a copy of the handlers on every iteration is a terrible idea.

I like the idea of using standard collection classes, but if that will cause the Video library to now throw an error when `dispose()` gets called, that isn't great either. Let me take another look at what's needed over there and perhaps make a PR there instead. At the moment there are problems with using multiple Movie objects or trying to shutdown a Movie object without also exiting the Sketch. The PR I made helps address that, but there are other ways to address those issues.

benfry:
Ok, just checked in a possible fix if you'd like to give it a shot.

hx2A:
Your approach for fixing this is better than what I came up with. Thank you.

I just tested this and I do get a null pointer exception. I think you made a typo in the first line of the `remove()` method. I believe you meant to write:

```
    public void remove(Object object) {
      if (removals == null) {
        entries.remove(object);
      } else {
        // Currently iterating the list of methods, remove this afterwards
        removals.add(object);
      }
    }
```

The checked in code has `if (removals != null) {` , which is the opposite of what you want.

Once I made that change the code worked perfectly. I can run multiple movies and dispose of them without exiting the sketch. The movie resources (gstreamer) are shut down properly, with no memory issues or exceptions thrown.

benfry:
Ha, whoops. Thanks for the catch.

hx2A:
You're welcome. :) We've all been there.

I just sent you an email about py5, we should talk about that when you have some time to spare.

Issue #214, opened 2021-06-25
Module javafx.base not found


letorbi:
## Description
Processing is not able to start after building it from the alpha 5 or the latest master code. 

## Expected Behavior
Well, processing should start without an exception ;)

## Current Behavior
Calling *build/linux/work/processing* returns with the following exception: `java.lang.module.FindException: Module javafx.base not found`

Here the full output of *build/linux/work/processing* with logging enabled:
```
APPDIR=/home/letorbi/Desktop/processing4/build/linux/work
DIR=/home/letorbi/Desktop/processing4/build/linux/work/java
L=rt.jar
LIB=
JDKCP=
SUCCESS=0
[...]

PR #215, opened 2021-06-25
Set correct --module-path in start script

letorbi:
This PR changes the `--module-path` parameter in the Linux start scripts from `modes/java/libraries/javafx/library/linux64` to `modes/java/libraries/javafx/library/linux64/modules`.

This prevents a `Module javafx.base not found` exception during the initialization of the boot layer. Details can be found in issue #214.

### DIFF
--- a/build/linux/processing
+++ b/build/linux/processing
@@ -113,5 +113,5 @@ else
   fi
   cd "$APPDIR"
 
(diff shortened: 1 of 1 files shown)


benfry:
Thanks!

PR #216, opened 2021-06-27
Fix font choice other than Source Code Pro

dzaima:
Choosing a font other than Source Code Pro in the options resulted in a variable-width fallback/default font being loaded in.

This was caused by the code effectively doing `new Font("editor.font.family", …)`, using the preference name variable, instead of the retrieved value.


benfry:
Thanks, I was looking for that one. Ironically, all those changes were kicked off by a fix for the same issue (the attr name being used rather than the font name) elsewhere in the code.

PR #219, opened 2021-07-03
Allow syntax error to try to complete text edits.

sampottinger:
Do not stop trying to generate java code when there is a syntax error but also have the completion generator check that types are available.


benfry:
Fingers crossed; thank you!

PR #220, opened 2021-07-05
Doclet changes

Anadroid:
This PR:

1. changes the subcategory names to not have `_ `but `space` or `&`
2. fixes category and category on `mask()` so it doesn't appear under `PImage` category on the main reference page (now its `image:pixels`)


Anadroid:
Sorry, this will first go to the DSI fork, and then once we transfer it to a new repo there will be a new PR.

PR #222, opened 2021-07-06
Doclet changes

Anadroid:
This PR:

1. removes the doclet from the `processing4` repo
2. changes subcategory names in java source files that had an `_` to have a `space` or `&` instead
3. fixes the category and subcategory for the `mask()` reference


benfry:
Thanks!

PR #225, opened 2021-07-14
fix PShapeOpenGL#transform() (fixes #217)

### LINKED ISSUE #217: PShape.scale not working with PShape.resetMatrix when P2D renderer is used
## Description

PShape.scale does not play well with PShape.resetMatrix when P2D renderer is selected.
## Expected Behavior

 I believe this to be a bug. I expected the ellipse to be drawn and scaled up absolutely by 2. I tried other svg's (loadShape) and the problem still occurred.
 Without P2D: 
image
With P2D:
image

## Current Behavior
 
The ellipse is not drawn at all. I think the issue lies within PShape.scale because PShape.translate seems to work fine, or possibly P2D and PShapeOpenGL.

## Steps to Reproduce

```processing
PShape s;

void setup(){
  size(1000,1000,P2D); // Removing P2D allows it to work.
    s = createShape(ELLIPSE,0,0,100,100);
}


void draw(){
  background(0);
  translate(width/2,height/2);
  // s.translate(-200,100); This works!
  s.scale(2);
  shape(s,0,0);
  s.resetMatrix();
} 
```

## Your Environment

 

* Processing version: 4.0a5
* Operating System and OS version: Windows 10 Build 19041.1052
* Other information:

cathiecode:
Quick fix for #217.

### DIFF
--- a/core/src/processing/opengl/PShapeOpenGL.java
+++ b/core/src/processing/opengl/PShapeOpenGL.java
@@ -1397,6 +1397,7 @@ public class PShapeOpenGL extends PShape {
       } else {
         transform.scale(args[0], args[1]);
         PGraphicsOpenGL.invScale((PMatrix2D)transformInv, args[0], args[1]);
+        tessellated = false; // Quick fix for https://github.com/processing/processing4/issues/217
       }
       break;
     case MATRIX:


benfry:
Thanks!

PR #229, opened 2021-07-24
Jssc fix libs and sync with upstream

### LINKED ISSUE #119: JSSC binaries for Windows appear to be macOS files
…or at least `codesign` thinks so. Is it possible that the cross-compilation for these libs wasn't set quite right? They're `.dylib` files after all.

For that matter, we can also nix the windows_32 versions.

sampottinger:
Updated JSSC including a sync to the upstream library's latest fixes and migrated to new cleaner native libs. Resolves #119 (finally). Tested on the following:

 - macOS 11.4
 - Windows 10
 - Elementary 5.1 (linux)


sampottinger:
Whoops... looks like intellij got up in there... one sec.

benfry:
Thanks, just some changes needed here with the name of the platform-specific subfolders. For instance, it needs to be `linux64` not `linux_64` to export properly. And we can leave out the 32-bit versions everywhere (except ARM, perhaps) since everything is 64-bit. So the folders should be more or less the same as before:

```
linux64 (64-bit Intel)
linux-arm64 (64-bit ARM for RPi)
linux-armv6hf (32-bit ARM for RPi)

either 
macosx (if it's a 32- and 64-bit fat binary)
or 
macosx64 (if it's just a 64-bit binary)

and then
macosx-arm64 (provisional, since we don't have support yet)
```

(This adds `macosx-arm64` and drops the unnecessary 32-bit versions.)

The macOS names will need to change someday, probably becoming `macos` instead of `macosx`, though the `-arm64` is confusing here for folks, so haven't decided what best to do there. (Confusing meaning that if we export an application and the folder is called `application.macosx-arm64` that most humans will be confused that it means “macOS on my M1 Mac”.

sampottinger:
Ah got it @benfry. I think these just reflect back what JSSC is doing upstream. I can change these targets.

sampottinger:
Have a draft but testing across platforms.

sampottinger:
@benfry, Actually ended up syncing again to master and they are bundling the natives in the jar now. See updated PR.

benfry:
Thanks! Now merged for beta 2.

PR #230, opened 2021-07-25
Allow preproc service to run on .java files.

sampottinger:
Great news! Because we have unified the preprocessing pipelines and everything runs through ANTLR the same way, we can drop `pdexEnabled`! Simple test...

java_test.pde:
```
void setup() {
  TestClass testClass = new TestClass();
  println(testClass.getVal());
}

void draw() {
}
```

TestClass.java:
```
public class TestClass {

  public String getVal() {
    return "val";
  }
  
}
```

### DIFF
--- a/java/src/processing/mode/java/Compiler.java
+++ b/java/src/processing/mode/java/Compiler.java
@@ -74,7 +74,6 @@ public class Compiler {
 
     String[] sourceFiles = Util.listFiles(build.getSrcFolder(), false, ".java");
     String[] command = PApplet.concat(baseCommand, sourceFiles);
-    //PApplet.println(command);
 
     try {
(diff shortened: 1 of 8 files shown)


sampottinger:
Just FYI...

Screen Shot 2021-07-24 at 6 15 20 PM

sampottinger:
Switched this to draft... realized that package statements could cause trouble...

sampottinger:
Sorry about that @benfry, this is resolved. The JDT needed a little bit of help but, after some testing, this seems performant and we have error reporting on ".java" tabs even if they are in their own packages.

sampottinger:
Setting back to regular PR. Ready for review when you are Ben!

benfry:
Woohoo! Thanks!!

PR #237, opened 2021-08-08
Add per-mode language bundles for i18 support in contributed modes

codeanticode:


jeremydouglass:
Would this be minimally documented somewhere so that mode authors / maintainers could know how to add such support?

codeanticode:
@jeremydouglass that's a great point. With the implementation proposed in this PR, the mode strings and any (optional) translations would be provided within the mode/languages folder, with the files mode.properties (english) and additional mode_<lan>.properties. For an example, see the android mode:

https://github.com/processing/processing-android/tree/master/mode/languages

As with the PDE, the translation will default to the English in the case  a particular string is missing in the properties of the current system language. Another detail about this implementation is that a string is missing altogether in the mode's language properties (both in the current system language and English), then it will try to find it in the PDE's properties. This would be useful for text in the UI of the mode that is unchanged from the base editor, so the mode can just rely on the base language properties.

Where this should be documented? I can provide an expanded version of these guidelines once the PR is merged and available in an upcoming Processing release.

jeremydouglass:
> Where this should be documented? I can provide an expanded version of these guidelines once the PR is merged and available in an upcoming Processing release.

In the past we've kept mode developer documentation of this kind on the Processing 3 github wiki. So, perhaps on the Processing 4 wiki? Honestly anywhere as long as we can point to it consistently, e.g. on the forums, and maybe leave a forward note on those P3 wiki pages pointing out that P4 is in a separate place.

benfry:
@codeanticode Do we really need to change the API to pass around the `Mode` instance? How about we prefix Mode-specific strings with `android.` or something similar? The “correct” method would be to use the full package name, but chances are `android.` and `python.` and others should be fine in practice. We just don't have that many Modes after all.

codeanticode:
I think one advantage of these changes, which has the drawback of changing the internal API as you pointed out, it that contribs can reuse strings in the base localization, since it's likely the keep many elements of the parent's UI (such its the case with modes). But I need to look at the code again to see if this is possible by using the package name.

benfry:
@codeanticode I don't think it needs to be full package name, but I was thinking you had an earlier patch that was simpler… You just need a string and if it's mode-specific, it should just have `android.` in front of it. 

Meaning that all the strings go into the same pile. So in fact, we could basically require that a Mode have its own strings prefixed (using `android.` or whatever), and if not, the load method can reject them so a Mode author doesn't inadvertently overwrite base strings.

I had mentioned using the full package as the prefix, which would be a more complete/more "correct" way of handling this, but we have so few Modes, it's dumb to be pedantic like that. I think `android.` and `jython.` would be fine.

benfry:
Implemented for 4.0 beta 5 based on the previous comment. Details in https://github.com/processing/processing4/commit/ddd4b1075398782bd1bdd854c16d2aa61fcc02da.

Issue #196, opened 2021-05-18
PShape performance could be improved using buffer object streaming 


codeanticode:
Performance was quite good with GLModel (from the [GLGraphics](http://glgraphics.sourceforge.net/) library) in certain instances where PShape still struggles (for example when modifying the coordinates of many vertices at once), because GLModel used a technique called [buffer object streaming](https://www.khronos.org/opengl/wiki/Buffer_Object_Streaming). This technique allows one to write data directly into a regular Java buffer, which internally is mapped to video memory. This relies on the function [glMapBufferRange](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glMapBufferRange.xhtml), which was available only on desktop, not on mobile (OpenGL ES), back when the OpenGL renderer in Processing 2 & 3 was implemented it. In order to keep code parity between desktop and mobile versions of the OpenGL renderer Processing, buffer streaming was not used in PShapeOpenGL. It turns out that glMapBufferRange eventually became available in [OpenGL ES 3.0](https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glMapBufferRange.xhtml). So it could be possible to re-implement buffer streaming to improve PShape performance, perhaps in combination with other PShape enhancements such as this one https://github.com/processing/processing/issues/2280, for both Processing 4 and new versions of the Android mode. [to be tagged as proposed enhancement]

benfry:
Since you're the maintainer of that code, are you offering to implement it? Or is this a request for someone else to do so?

codeanticode:
I'd like to implement it. Just could not assign issues to myself in this repo :-)

benfry:
That's odd, it should have similar permissions as the 3.x repo, but let me know if I need to set something for you.

codeanticode:
Looks like I'm able only to unassign myself from the issue, but that's ok.  You can add "enhancement" and "opengl" as labels.

codeanticode:
Explored stream object copy, actually not that hard to implement: https://github.com/codeanticode/processing4/commit/a23aaa51980e0bfc1dc2dc7e3c1c47ba3ff8648f

That could help, but what makes PShape sluggish compared with the old GLModel class is that the latter operated directly on the tessellated geometry that's rendered through GL. PShape adds a lot of overhead necessary to properly tessellate primitives, arbitrary polys, stroke lines, etc. 

[...]

PR #238, opened 2021-08-08
Implement buffer object streaming and finalize attribute API in PShape

codeanticode:


micycle1:
What's the outcome of these changes? Better performance with the OPENGL renderer?

codeanticode:
@micycle1 Yes, the rationale is explained in this issue: https://github.com/processing/processing4/issues/196
By allowing the editing of tessellated geometry without re-tessellation, you could get significant performance improvements, at least in certain use cases (dynamically-changing shapes). None of this affects the "immediate mode" (i.e.: re-drawing the geometry in each frame).

codeanticode:
Reopened here: https://github.com/processing/processing4/pull/314

Issue #240, opened 2021-08-09
Imports whose name contains "color" cause Syntax error


micycle1:
## Description
Imports whose name contains "color" cause a syntax error since Processing interprets these as the `color` datatype.

![image](https://user-images.githubusercontent.com/9304234/128710835-9819e00e-875f-4ed3-9522-402b2541d233.png)

## Your Environment
* Processing version: 4.0 Beta 1
* Operating System and OS version: Windows 10

benfry:
Hm, that's unfortunate, though I'm not sure it's a bug—does that work with `int`, `float`, or any other primitive types? `color` is what's considered a “reserved word” in the language.

(Edited to add: if it was working in 3.x, that was probably a bug in the preprocessor, unfortunately.)

micycle1:
It was working in 3.x.

![image](https://user-images.githubusercontent.com/9304234/129017193-31a699f4-6451-4a9f-bef8-734c37594bff.png)

![image](https://user-images.githubusercontent.com/9304234/129017265-fd5e2c7b-1dd6-4ae7-927c-4730fb2cadc5.png)

sampottinger:
I’ll look into this right away @micycle1. Thank you for reporting it.

sampottinger:
Apologies @benfry and @micycle1 for not having seeing this during the alpha testing. Color is a little strange because, gramatically, it shouldn't be able to appear in a fully qualified name (or import) like int cannot. However, I would hate to have any incompatibility. @benfry, I'm suggesting a change to the grammar. Please see #246.

PR #246, opened 2021-08-11
Fix color in appearing in fully qualified names.

sampottinger:
Addressing #240, allow color to appear in fully qualified names like as part of import statements. It's not the cleanest solution (other primitive types could not appear here) but I don't see too many alternatives.

### DIFF
--- a/java/src/processing/mode/java/preproc/Processing.g4
+++ b/java/src/processing/mode/java/preproc/Processing.g4
@@ -4,7 +4,8 @@
  *	- changes main entry point to reflect sketch types 'static' | 'active'
  *	- adds support for type converter functions like "int()"
  *	- adds pseudo primitive type "color"
- *	- adds HTML hex notation with hash symbol: #ff5522 
+ *	- adds HTML hex notation with hash symbol: #ff5522
+ *  - allow color to appear as part of qualified names (like in imports)
  */
 
 grammar Processing;
@@ -47,8 +48,8 @@ variableDeclaratorId
 // https://github.com/processing/processing/issues/93
 // prevent from types being used as variable names
 warnTypeAsVariableName
-    :   primitiveType ('[' ']')* { 
-            notifyErrorListeners("Type names are not allowed as variable names: "+$primitiveType.text); 
+    :   primitiveType ('[' ']')* {
+            notifyErrorListeners("Type names are not allowed as variable names: "+$primitiveType.text);
         }
     ;
 
@@ -89,6 +90,10 @@ colorPrimitiveType
     :   'color'
     ;
 
+qualifiedName
+    : (IDENTIFIER | colorPrimitiveType) ('.' (IDENTIFIER | colorPrimitiveType))*
+    ;
+
 // added HexColorLiteral
 literal
     : integerLiteral
@@ -127,4 +132,3 @@ LINE_COMMENT
     ;
 
 CHAR_LITERAL:       '\'' (~['\\\r\n] | EscapeSequence)* '\''; // A bit nasty but let JDT tackle invalid chars
-
--- a/java/src/processing/mode/java/preproc/Processing.g4
+++ b/java/src/processing/mode/java/preproc/Processing.g4
@@ -5,7 +5,7 @@
  *	- adds support for type converter functions like "int()"
  *	- adds pseudo primitive type "color"
  *	- adds HTML hex notation with hash symbol: #ff5522
- *  - allow color to appear as part of qualified names (like in imports)
+ *      - allow color to appear as part of qualified names (like in imports)
  */
 
 grammar Processing;
--- a/java/src/processing/mode/java/preproc/Processing.g4
+++ b/java/src/processing/mode/java/preproc/Processing.g4
@@ -5,7 +5,7 @@
  *	- adds support for type converter functions like "int()"
  *	- adds pseudo primitive type "color"
  *	- adds HTML hex notation with hash symbol: #ff5522
- *      - allow color to appear as part of qualified names (like in imports)
+ *     - allow color to appear as part of qualified names (like in imports)
  */
 
 grammar Processing;
--- a/java/test/processing/mode/java/ParserTests.java
+++ b/java/test/processing/mode/java/ParserTests.java
@@ -385,4 +385,9 @@ public class ParserTests {
     expectGood("smoothparamstatic");
   }
 
+  @Test
+  public void testColorInImport() {
+    expectGood("colorimport");
+  }
+
 }
--- a/java/test/resources/colorimport.expected
+++ b/java/test/resources/colorimport.expected
@@ -3,6 +3,8 @@ import processing.data.*;
 import processing.event.*;
 import processing.opengl.*;
 
+import test.color;
+
 import java.util.HashMap;
 import java.util.ArrayList;
 import java.io.File;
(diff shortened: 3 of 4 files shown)


dzaima:
Other primitive types aren't allowed to be part of import statements anyways. A similar problem is with `var`, which is a valid identifier everywhere except for types (which this grammar doesn't support either). The "proper" way would be for `'color'` and `'var'` to not be special tokens, but instead specially handled identifiers in types (and I guess in the `color()` function), but that's certainly much more complicated.

Edit: Does the grammar even need to know about `var`? Shouldn't it being passed through as any other identifier work just fine?

benfry:
Yeah, I don't think this is actually something we want to enable/provide a hack to make it work… Using a reserved word is a no-no in any language, and carving out separate behavior is a maintenance nightmare. It's a bummer, but I think the alternative is worse.

sampottinger:
Respect your decision either way Ben. My only thought is that it does make us unable to use some Jars since color isn’t reserved in Java. However, I think the concern around grammar integrity is also valid.

All that in mind, I checked and one alternative we do have for folks is that they can still use .java files to import things with color in their qualified names which they can then reference in the .PDE files.

benfry:
I think unless it's possible to have `int` or other primitive types in a package name, we shouldn't carve out an exception in the grammar. (If that's confirmed, we can close this.) We've spent so much time tracking down little quirks whenever we do something nonstandard like this that I don't want to introduce another exception.

There are also a lot of alternatives:

* Set `preproc.color_datatype = false` in `preferences.txt`
* Use a `.java` tab for the import (which won't use the preprocessor)
* Change the name of the package
* Use a full-fledged development environment like IntelliJ or Eclipse

The bottom line is that `color` is a primitive data type, and just an alias for `int`, and should behave just like `int` does in Java. It exists because it's super weird to explain to beginners that `int` is actually a color (sometimes but not always), and there's a significant performance impact if all colors are defined using objects.

jeremydouglass:
This also means the color part of java.awt will not be directly accessible, correct?

```
import java.awt.color.ColorSpace
```

https://docs.oracle.com/en/java/javase/11/docs/api/java.desktop/java/awt/color/package-summary.html

If `import foo.color` worked in Processing 3 and now will not, I do wonder how many libraries are impacted. I did a quick scan of ~130 repos that were locally cached from the Archive for Processing collection, and I did not find any `import *.color` examples -- the closest that I found was Color classes, e.g. in colorscheme: https://github.com/josephtaylor/ColorScheme/blob/master/src/main/java/jto/colorscheme/Color.java

That said, there are maybe 500 packages I could check in the index -- and that doesn't speak to the larger java ecosystem.

jeremydouglass:
p.s. I should have also mentioned -- micycle's example in the issue (micycle.pgs.color.*) from the [PGS](https://github.com/micycle1/PGS) library was not found my scan because that library was published in May, and I hadn't scraped it yet.

sampottinger:
Just point of info, no int isn't allowed in Java package names. https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html

jeremydouglass:
There is [no other package in Java 11](https://docs.oracle.com/en/java/javase/11/docs/api/allpackages-index.html) with a `.color` namespace other than `java.awt.color.ColorSpace`

The only other major library I found out there is in Apache Commons -- [org.apache.commons.imaging.color](https://commons.apache.org/proper/commons-imaging/apidocs/org/apache/commons/imaging/color/package-summary.html).

[Guava's packages](https://guava.dev/releases/30.1-android/api/docs/allpackages-index.html) do not have a `.color` namespace.

sampottinger:
Yeah I think I see your point @jeremydouglass…. Especially for awt.color.

I know it’s weird to carve out an exception for color but it is the only primitive with this issue and the grammar change doesn’t concern me too much. It shouldn’t get caught up in the preprocessing rewrites because of the context and there’s tests for preproc in particular that could help catch a regression. I do see how this could be confusing for a new developer so I’d still be comfortable with the carve out.

jeremydouglass:
That sounds good to me. As it stands the consequences are:

1. you can't use `java.awt.color.ColorSpace`
2. you can't use `org.apache.commons.imaging.color`
3. micycle would need to renamespace the color part of PGS to run on Processing 4
4. there might be something out there we didn't find ~~but probably nothing major~~ still hard to say how much
5. it may come up periodically in the future as a weird corner case for new Processing or Java libraries
6. **Edit** you can't import `toxi.color`

Those aren't great, but ~~they are relatively minor compared to what I feared we might find~~ I'm not sure how to rate it in bad-ness. If there was a long list of legacy libraries that would be broken by this then I would be begging for a grammar carve-out to allow imports with color namespaces. As it is, I *still* think a grammar carve-out might still be nice, BUT this isn't a hill I'm trying to die on if Ben disagrees. ;)

postspectacular:
FWIW I just got informed about this issue by a toxiclibs user, i.e. that `import toxi.color.*;` doesn't work in v4 anymore, but did work in any other recent & previous version in the last 20 years...

As you know, I don't actively maintain the library anymore, but if the final answer really is as @benfry stated above, then I will try to cut a new release with breaking changes (i.e. the package name changed) and dozens of examples sketches updated...

IMVHO this issue shouldn't even be specifically about `color`, but usage of 'reserved' words within fully qualified names in general. Inarguably, there's a clear syntactic, semantic and practical difference between using a reserved word like `color` as standalone individual token (i.e. as actual keyword/type name) or the same word occurring in a syntactically correct manner *as part of a fully qualified package name*... These two usages cannot just be equated (not should they be) and I literally can't think of any other language/environment where these two distinct use cases are treated as equal and/or as proposed here. But again, am not here to register my opposition, just giving my 2 euro cents... ✌️

dzaima:
> I literally can't think of any other language/environment where these two distinct use cases are treated as equal and/or as proposed here.

@postspectacular Java is one such language with `var` (e.g. `var var = 2;` defines an integer variable named `var`; IIRC there's some more stuff like this in Java). "context sensitive keywords" is the search term for this stuff. They're used out of necessity for backwards compatibility of course, so it's not a thing you _should_ do, but it's a reasonable option if you do want the backwards compatibility. (or, in this case, compatibility with Java too)

postspectacular:
@dzaima It shows then, I haven't used Java for ~10 years... Could have sworn there wasn't such a restriction for package names (and those are the only ones which matter here), but I stand corrected. The `var` example you've given I'd have classified as reserved-word-used-as-individual-identifier (of course illegal), but IMHO v. different from a _multipart_, dot-separated package name...

Anyhow, the only other thing somewhat bugging me a little here is that after 2 decades (congrats, btw!) of established behavior in P5, this use of `color` et al is now being counted as a 'illegal' and the new behavior marked as 'working as intended'. I do kind of get it, however, this stuff is breaking hundreds of existing projects (not just my own, also lots of students/peers) which otherwise would still work and I'm now having to explain this (again) to users... I shall file it under the same category as when years ago P5 suddenly stopped auto-importing various `java.*` packages, triggering a flood of confused users and requiring outside effort (i.e. not P5 core team) to spend extra effort updating existing workshop examples/projects/documentation. I do get (some of) the reasoning, but IMHO those kind of changes should be communicated more clearly/prominently (e.g. via better error messages, maybe w/ a link to a GH issue), not just a sudden generic syntax error. Again, my 2 cents...

benfry:
Given the impact as outlined here https://github.com/processing/processing4/pull/246#issuecomment-907671906 that changes the priorities quite a bit. If @sampottinger has a fix that's working without too much impact, it sounds like we should make use of that. Sam, can you verify that you're comfortable with the reliability of the fix?

benfry:
Giving this a shot for 4.0 beta 2 for sake of compatibility. Fingers crossed…

Issue #48, opened 2020-01-18
Update graphics and visual design elements for 4.x


benfry:
Things may still say Processing 3, and use its color scheme. We'll be updating that before the release.

urbanskimichal:
Is it already designed and needs to be implemented or this task is about proposing a design? If the latter - what elements exactly?

benfry:
This is just a note that they'll need to be swapped out before launch.

As for the design, it's the one thing that *does not* actually need additional volunteer help.

pr1metine:
Hi guys,
Are you talking about the PDE? In that case, I'd like to propose a complete overhaul of it. Maybe we replace it with a [Language Server](https://microsoft.github.io/language-server-protocol) Plugin e. g. for [VS Code](https://github.com/microsoft/vscode) or ship a [Theia](https://github.com/theia-ide/theia) application with Processing.

benfry:
That's a separate discussion, see here: https://github.com/processing/processing/issues/5423

benfry:
(Locking this since the issue can be read too broadly and to avoid this turning into a general discussion about the UI of the PDE.)

benfry:
4.0 alpha 6 introduces a new color set *that is not final*. You can also set your own colors and styles [here](https://github.com/processing/processing4/wiki/Themes). We're working on ways to make theming more personal and customizable (without having to mess with `theme.txt`), but more on that later.

benfry:
4.0 beta 3 lands the first (major) round of the theming changes. Read more about it in the release notes or at https://github.com/processing/processing4/wiki/Themes.

It also has an updated application icon, which is better than the previous but:
* May not yet be final – it's a version of the official logo that's been tweaked for the small size and icon usage
* Is using the “Big Sur” style, even for Windows and Linux releases – not ideal, but also unlikely that we'll have time to maintain separate icons for each platform. I don't personally like the Big Sur style, but it's probably the best of several bad options.

Themes are still incomplete: 
* The editor “tokens” (what colors text in code) are not yet implemented properly relative to the overall theme. They're just using the old defaults.
* We'd like to somehow incorporate the generative theme engine we're using for the set of 16 into the software itself, so that folks can pick their own starting colors, accents, and styles.
* The Editor Console needs a custom scroll bar
* The Contribution Manager hasn't been redone
* The debug window hasn't been touched

[...]

PR #260, opened 2021-08-26
Changed macOS app icon to Big Sur style and replaced generic file icon with macOS autogenerated icon for macOS Catalina and Big Sur

andylin2004:
This is a PR to just replace the macOS icon to the Big Sur style so that it fits in with all the other icons instead of a sharp, jagged square. The document icon has also changed in macOS 10.15+ to show the app icon, and on macOS 11, the file extension. All previous versions of macOS will continue to show the generic file icon, as shown [here](https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/document-icons/).


benfry:
Thanks, though the icons aren't finalized yet. That's being tracked here: https://github.com/processing/processing4/issues/48

PR #274, opened 2021-09-27
Fix welcome screen dismissing

### LINKED ISSUE #253: Startup message is always shown on startup
## Description

It seems that the startup message is always shown on startup. The checkbox should also be checked since it's written "Show this message on startup" but enabling it and disabling it doesn't change anything.

Screenshot from 2021-08-19 12-12-27

## Expected Behavior

The checkbox should be checked and unchecking it should hide the startup message.

## Steps to Reproduce

1. Download Processing 4
2. Run `install.sh`
3. Open Processing 4

## Your Environment

 

* Processing version: `4.0b1`
* Operating System and OS version: `Pop!_OS 21.04`

dzaima:
Fixes #253 by updating the couple leftover old config reads & writes to reflect the renaming of the value.


benfry:
Thanks for the catch.

PR #286, opened 2021-10-05
Enable JVM Font Antialiasing for Linux

pavlus:
Should add smoothness for fonts in the Processing GUI most of the time.
With change: 
image
Without change:
image

(GitHub compressed images, but the difference is still visible)

### DIFF
--- a/build/linux/processing
+++ b/build/linux/processing
@@ -114,5 +114,5 @@ else
   cd "$APPDIR"
 
   #java -splash:lib/about-1x.png -Djna.nosys=true -Djava.library.path=lib:modes/java/libraries/javafx/library/linux64 --module-path=modes/java/libraries/javafx/library/linux64/modules --add-modules=javafx.base,javafx.controls,javafx.fxml,javafx.graphics,javafx.media,javafx.swing,javafx.web -Xmx512m processing.app.Base "$SKETCH" &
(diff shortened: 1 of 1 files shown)


benfry:
This shouldn't be necessary, anti-aliasing should be turned on by default in the Java VM. Is your desktop manager turning it off? What sort of display are you using? 

The option that you're using here (`lcd`) specifically overrides anti-aliasing so that it works for LCD screens that use RGB bars in a horizontal orientation, which is (very?) common, but not universal. If I set this as the default, it will break smoothing on other devices, or give it strange halos because the pixels are in a different order or vertical instead of horizontal. 

Docs for the flag [here](https://docs.oracle.com/javase/7/docs/technotes/guides/2d/flags.html#aaFonts) and the `lcd` option maps to [this setting](https://docs.oracle.com/javase/7/docs/api/java/awt/RenderingHints.html#VALUE_TEXT_ANTIALIAS_LCD_HRGB).

pavlus:
It's not enabled by default, unless turned on via XSETTINGS properties `gnome.Xft/Antialias`, or `fontconfig/Antialias`, which can be set by desktop manager.

See: https://github.com/openjdk/jdk/blob/6765f902505fbdd02f25b599f942437cd805cad1/src/java.desktop/unix/classes/sun/awt/UNIXToolkit.java#L331-L344

benfry:
See the links I posted: the option you specify overrides the display to use HRGB ordering for all cases, meaning that displays using HBGR, VRGB, VBGR, etc will all look worse than before.

PR #288, opened 2021-10-10
Detect if calling special methods on PApplet or not (and restore unit tests).

### LINKED ISSUE #272: error when calling smooth() on PGraphics
## Description

When attempting to set `smooth()` on a PGraphics instance an error is thrown.  

## Expected Behavior

The processing JavaDocs show that a PGraphics instance can receive `smooth()`

## Current Behavior

Calling `smooth()` on a `PGraphics` instance throws the following error:

```
Syntax error on token ".", invalid VariableDeclarator
```

## Steps to Reproduce

1. Run the following code 
```
PGraphics gfx;

void setup() {
 
  gfx = createGraphics(width, height);
  gfx.smooth();
}

void draw() {
}
```
2. Behold the error

## Your Environment

 

* Processing version: Processing 4.0b1
* Operating System and OS version: Mac OS X 11.5.1 (20G80)
* Other information:  iMac (Retina 5K, 27-inch, 2019) Radeon Pro 570X 4 GB

`smooth()` works fine when applied to the top level of the app and seems to apply itself to subsequent `PGraphics` within.  There is some mention that in Processing 3+ that `smooth()` can only be run once during setup.  Perhaps there is a usage issue with `smooth()` and that it should be removed from the `PGraphics` spec.

### LINKED ISSUE #287: Mockito needs to move to new version
We are on Mockito 1.x but it looks like Java 17 finally broke it. Getting to Mockito 4 might take a little work, including getting a fat jar. Having the tests runnable though is important for preproc work.

sampottinger:
Resolves #287 and resolves #272.


sampottinger:
@benfry it looks like a lot of files but I just had to update stuff for new JUnit / Mockito. The only non-test modifications are at: https://github.com/processing/processing4/pull/288/files#diff-e99bcf0e145623e4fd8cf2d354875b44efc1161ee64e41ebf53714060e94c60e.

That said, I did add a test for this case both with and without params while also confirming that previous prepoc tests are still passing.

benfry:
Thanks!

PR #298, opened 2021-10-24
Workaround for exceptions when resizing AWT windows while sketch is running

keithcbentley:
Change PSurfaceAWT to use atomic boolean instead of being synchronized.

This fixes a timing based deadlock when resizing a sketch as described in https://github.com/processing/processing/issues/5579.


benfry:
This should be resolved in 4.0 beta 5.

PR #301, opened 2021-10-30
Update Ukrainian language strings

rzats:
Fills in a couple of missing localization strings, and removes unused ones from the movie maker rework.


benfry:
Awesome, thank you!

PR #329, opened 2022-01-01
Set Processing Icon in Splash

### LINKED ISSUE #297: Splash screen has default OpenJDK icon
## Description/current behavior

The splash screen shows up like this in the Windows task bar:

image

This appears to be the OpenJDK icon.

## Expected Behavior

The splash screen should either not appear on the task bar at all (Processing 3 has worked this way), or use the Processing icon.

## Steps to Reproduce

Start `processing.exe`.

## Your Environment

Freshly downloaded 4.0 beta 2 release of Processing.

Windows 10 21H1.

The task bar entry corresponds to `javaw.exe`, which also has this icon in Windows Explorer. Ideas include replacing the icon resource in javaw.exe at build time, or using some Windows mechanism to supply an alternative icon. Unsure whether this conflicts with any licenses though.

Not showing the splash screen in the task bar would be the best option I guess.

rupeshkumar22:
Fixes #297

Verified taskbar icon after build and run, now it is showing processing icon during splash window.

### DIFF
--- a/app/src/processing/app/ui/Splash.java
+++ b/app/src/processing/app/ui/Splash.java
@@ -24,6 +24,10 @@ public class Splash extends JFrame {
 
 
   private Splash(File imageFile, boolean hidpi) {
+    
+    // change default java window icon to processing icon
+    processing.app.ui.Toolkit.setIcon(this);
+    
     this.image =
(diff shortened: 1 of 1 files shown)


benfry:
Great; thanks!

rupeshkumar22:
@benfry I think the same, using try/catch would be a better case.

PR #360, opened 2022-01-21
Portable settings

Eskaan:
## The Feature
Added a "settingsPath.txt" file inside of the App itself that, when a path is put in, uses the given path to redirect the settings from the usual %AppData%/Processing folder to the given folder.

## What it does
It modifies the `getSettingsFolder` method in `Base` that provides the folder so that it first checks (and creates) the settingsPath file in the app folder. If this file then contains a line, it will try using this String as the setting folder (relative paths possible). If the file is empty or doesn't exist, it will use the default directory for the settings. 

## Sidenotes
- I know you don't want too much code in the Base class (you mentioned it somewhere) but I see this as the best solution.
- If you take the windows and linux versions of the app and edit each app's setting paths it gets fully portable across systems :D

## Draft
- [ ] Will follow up whit GUI implementation soon. Under the sketch location or at the bottom?
- [ ] Maybe a parser for the relative paths to make them less confusing: 
`/home/apps/Processing.app/../../foo/bar/` -> `/home/foo/bar/`
- [ ] Check if relative path validation works on windows. If not.. Well a few if statements will do.
- [ ] Generally test feature on Windows since I am on a Mac.

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -27,6 +27,7 @@ import java.awt.*;
 import java.awt.event.ActionListener;
 import java.io.*;
 import java.lang.reflect.InvocationTargetException;
+import java.nio.charset.StandardCharsets;
 import java.text.SimpleDateFormat;
 import java.util.*;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


benfry:
Adding this back for 4.0 beta 4… Not unlike the approach here, you'll just need to make a `settings.txt` file inside `lib`, and include `settings.path=/an/absolute/path/for/settings` to make it work.

(The commit isn't pushed here yet because I'm breaking some other things at the moment and don't want to break the repo.)

benfry:
Those changes are now committed; if you run into problems, file an issue and I can look into it. I haven't had time to test it yet.

PR #381, opened 2022-01-29
Add ctrl and alt deletion

Eskaan:
Added the option to delete whole lines or just words by holding ctrl (on mac cmd) or alt while pressing the backspace key. Dosen't need much explanation.

Also fixed Autocompletion not including classes not being in the processing package. On top of that, classes had wrong names.

### DIFF
--- a/java/src/processing/mode/java/CompletionGenerator.java
+++ b/java/src/processing/mode/java/CompletionGenerator.java
@@ -1300,18 +1300,11 @@ public class CompletionGenerator {
 
     if (isImported) return false;
 
-    if (impName.startsWith("processing")) {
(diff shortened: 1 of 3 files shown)


benfry:
Thanks, though these need to be separate PRs. 

For the backspace behavior, you're looking for option-delete, which we implement per Apple's guidelines: https://support.apple.com/en-us/HT201236 Same for cmd-left moving through the line (or selecting it if shift is down).

Eskaan:
I tried to mimic Apples Option(alt)-Delete behavior, I just marked dots, commas and all opening brackets as word-separation characters. If I would cherry-pick the commit to another branch, would it be good?

Issue #275, opened 2021-09-27
How to develop P4 with intelliJ


clankill3r:
I would like to fix an issue with the current pdf exporter.
For this it would be great if I can edit processing and the pdf library from Eclipse or IntelliJ.
Since the repository comes with an `.idea` folder I suspect that more people are using IntelliJ for the development of processing.

I would like to know, how can I make changes and test in IntelliJ with a debugger?
I tried making a configuration, but I can't figure it out.

benfry:
I'm only just getting used to IntelliJ myself, so I don't have a good answer. At the worst, it's entirely possible to do it with just `ant run` ala the build instructions.

oseph:
I'm no IntelliJ expert myself, but here's how I've got processing4 setup in IntelliJ on macOS:

1. Clone the repository and build the project
2. Open the project folder in IntelliJ
3. Open Project Setting (`command/ctrl+;`)and select Modules in the sidebar.
4. Select all the modules and delete them with the `-` button
5. Click the `+` to `import modules` and select the core folder, and click yes to all the prompts.
6. You should be good to go!

keithcbentley:
I've attached a Word doc with detailed instructions.
[Processing on IntelliJ IDEA.docx](https://github.com/processing/processing4/files/7406025/Processing.on.IntelliJ.IDEA.docx)

codeanticode:
@benfry maybe the word doc contributed by @keithcbentley can be used as the basis for a section in the wiki?

PR #382, opened 2022-01-29
Update IntelliJ config files

Eskaan:
...and added IntelliJ build instructions.

IntelliJ should now select the right ant build file. I also tried to edit the style config for the formatter to match the project's Style Guidlines. IntelliJ before tried to find a Library named "jogl" but it was named "jogl-all" - so IntelliJ failed to find it. I also had to add the "gluegen-rt" Library to the librarys for the `com.jogamp.common` package.


Eskaan:
Fixes https://github.com/processing/processing4/issues/275

java/src/processing/mode/java/CompletionGenerator.java
```diff
@@ -1693,6 +1693,7 @@ static private void log(Object object) {
 
     // If the parsed code contains pde enhancements, take 'em out.
     // TODO: test this
+    // The Processing Foundation: Running untested code since 2016
```
benfry: What's the point of insulting the volunteers working on the project?

Eskaan: Really not meant as an insult, just stumbled upon this little Easter egg

PR #384, opened 2022-01-30
Preproc bug fixes and improvements

### LINKED ISSUE #290: mixing active and static mode throws the "wrong" error
## Description
In previous versions of Processing, when mixing active and static modes the error Processing would throw was:

**It looks like you're mixing "active" and "static" modes.**

In Processing 4, the error becomes a lot more cryptic:

**Syntax Error - Missing operator, semicolon, or '}' near 'setup'?**

As this is a mistake a lot of first learners make, it would be really helpful to get the old error message back.

## Steps to Reproduce
```
size(100, 100);

void setup() {
  
}
```

### LINKED ISSUE #317: Problem with function size(int arg, int arg) in Class
When the function `void size(int arg, int arg)` is used with two arguments in `Class` that's cause a syntax problem when this function is call directly in `void setup()`

Processing 4.0.2b / OS Monterey

Code to reproduce:
```
Truc truc = new Truc();
void setup() {
  size(200,200);
  truc.size(1,1); // problem >>> error à "."
  // func();
}

void draw() {
  truc.size(1,1); // no problem
}

void func() {
  truc.size(1,1); // no problem
}

class Truc {
  void size(int x, int y) {
  }
}
```

### LINKED ISSUE #371: Missing support for multi-line string text blocks
Java versions 15+ support multi-line strings in the form of text blocks delineated by three double-quotes:

String foo = """
some text
some more text
""";

The linux 4b3 install includes openjdk 17.0.1 2021-10-19, but the above code still generates a syntax error in the IDE.

Text blocks would be a very convenient way to include small shaders without having to resort to a separate external editor, though, granted, they won't have any syntax highlighting or automated formatting.

sampottinger:
Takes care of a few things in preproc:

 - Closes #290: Detects **mixed modes** in preproc by making a rule for it in ANTLR.
 - Closes #317: This was previously resolved but adding a unit test to make sure there isn't a regression regarding calls to **user defined size function**.
 - Closes #371: Added support for **multi-line strings**.

Two things to note:
 - TODO statements are added in where we need to switch to Java 17 (even though we have 17, we are still compiling for 11). Changing this now requires some changes including new JDT jars which I didn't want to do until we know for sure we are going with 17. Until then (and to support possible regression to Java 11), there is preproc rewrite support for multi-line strings. This means that the ANTLR preproc enables us to use multi-line strings in Java 11.
 - There's another piece of housekeeping here... While it has been reduced in scope, `SourceUtils` still has been hanging on. It has an internal automaton for scrubbing which will be difficult to retrofit for multi-line strings but its `processAST` and `scrubCommentsAndStrings` are redundant to what ANTLR is doing anyway so added a flag to disable them.


build/shared/lib/languages/PDE.properties
```diff
@@ -415,6 +415,7 @@ editor.status.bad.generic = Possibly missing type in generic near '%s'?
 editor.status.bad.identifier = Bad identifier? Did you forget a variable or start an identifier with digits near '%s'?
 editor.status.bad.parameter = Error on parameter or method declaration near '%s'?
 editor.status.bad.import = Import not allowed here.
+editor.status.bad.mixed_mode = You may be mixing active and static modes which is not allowed.
```
benfry: Let's use the old language for now, dropping the "which is not allowed." I'll update later with a link to the reference that explains what's going on, but the descriptions have unfortunately been dropped from the reference and I'm working on that with Casey.

java/src/processing/mode/java/preproc/Processing.g4
```diff
@@ -23,21 +23,29 @@ processingSketch
     :   javaProcessingSketch
     |   staticProcessingSketch
     |   activeProcessingSketch
+		|   warnMixedModes
```
benfry: Tabs sneaking in here…

java/src/processing/mode/java/preproc/Processing.g4
```diff
@@ -23,21 +23,29 @@ processingSketch
...
+// No method declarations, just statements
 staticProcessingSketch
     :   (importDeclaration | blockStatement)* EOF
     ;
 
 // active mode, has function definitions
 activeProcessingSketch
-	:	(importDeclaration | classBodyDeclaration)* EOF
-	;
+	  :	  (importDeclaration | classBodyDeclaration)* EOF
```
benfry: More tabs…

benfry:
Awesome, thanks. Added a few notes.

build/shared/lib/languages/PDE.properties
sampottinger: Resolved

java/src/processing/mode/java/preproc/Processing.g4
sampottinger: Darn it. Thanks! I’ll standardize these to soft.

java/src/processing/mode/java/preproc/Processing.g4
sampottinger: Same… sorry about that will standardize.

java/src/processing/mode/java/preproc/Processing.g4
benfry: No problem; looks like the file has it mixed which makes it tricky. As soon as these are set I can merge.

sampottinger:
@benfry - cleaned up the mixed spaces / tabs and standardized the spacing in `Processing.g4`.

benfry:
Awesome, thanks! In for 4.0 beta 5.

PR #400, opened 2022-02-10
Fix greedy multi-line string rule.

### LINKED ISSUE #398: multi-line strings do not terminate properly
If you attempt to create a pair of multi-line strings, they are both merged into one string:

```
String foo = """
hi
bye
""";

String bar = """
wee
woo
""";

println( foo );

```
generates

```
hi
bye
""";

String bar = """
wee
woo

```

and the variable "bar" does not exist.

The expected behavior is that foo and bar should each be a double-line string.  "foo" should terminate with the """ line immediately below "bye".  Multi-line strings should terminate at the first line that contains """.

sampottinger:
Single character fix which adds test for and closes #398 by making the rule non-greedy within ANTLR. Sorry about that @benfry.


benfry:
Thanks for the fix!

PR #422, opened 2022-02-24
Draft: Implement NURBSs

arnoudvanderleer:
In `todo.txt`, it says that it would be nice to have NURBSs implemented at some point. I know that it is a far-future goal, but it was fun to implement.

I am not very familiar with the processing way of doing things. Does this match the calling convention etc. of processing? Are there things that should be changed?

Usage:
```
size(400, 400);

NURBS n = new NURBS(
  new PVector[]{
    new PVector(300, 200),
    new PVector(300, 300),
    new PVector(200, 300),
    new PVector(100, 300),
    new PVector(100, 200),
    new PVector(100, 100),
    new PVector(200, 100),
    new PVector(300, 100),
    new PVector(300, 200)
  },
  new float[] {
    2, sqrt(2), 2, sqrt(2), 2, sqrt(2), 2, sqrt(2), 2
  },
  new float[] {
    0, 0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 4
  }
);

n.draw(g, 1000); // The number is the number of steps to subdivide into.
```


arnoudvanderleer:
I know that my code still needs comments. If the rest of my code is approved, I will add them. But what would be the preferred way of using a NURBS? Via the constructor+method shown above?

Adding a method to `PApplet` called `NURBS(x0, y0, z0, weight0, ..., weightn, knot0, knot1, ...)`, like the method for bezier curves, sounds suboptimal. This is because of the additional required inputs (weights, knot vector), as well as the fact that part of the usefulness of NURBSs comes from the possible variation in the number of points and length of the knot vector.

benfry:
Thanks @arnoudvanderleer! That was a very old comment in there (maybe 15 or 20 years) that I should probably remove. (I just haven't had time to go through every line of that file to clean it up.) 

At this point, it would be great to have this available as a Library to be used with Processing. After the feature has been used for a while by users (and debugged further, and you've received comments on its API), we can see about incorporating the code into Processing itself. This has been our approach for adding anything that's not a single line function. As you note, the additional inputs are what make NURBS useful, suggesting that a NURBS class would be a better approach.

arnoudvanderleer:
I will add this as a library at some point

benfry:
Great! Looks like you were able to make some good progress on it.

PR #433, opened 2022-03-03
Scale by pixel density to coordinates are consistent

### LINKED ISSUE #429: filter() incompatible with high pixel density on P2D and P3D
## Description

Using `filter()` (for example, `filter(GRAY)`, `filter(INVERT)`) in combination with a higher pixel density (set for example with `pixelDensity(displayDensity())`) behaves weirdly when using P2D or P3D.

## Expected Behavior

The filter would be applied with P2D/P3D and high pixel density in the same way as with the default renderer or with default pixel density.

## Current Behavior

The frame is drawn zoomed in on the top left corner. If the default renderer is used, everything works as expected. If no higher pixel density is used, everything works as expected. Interestingly, if we use `saveFrame()`, the saved image looks fine even though it is rendered wrong in Processing. Therefore, I'm attaching here screenshots instead of saved frames.

Default renderer or default pixel density:
Captura de pantalla 2022-03-02 a las 12 37 44

High pixel density and P2D/P3D:
Captura de pantalla 2022-03-02 a las 12 37 31

## Steps to Reproduce

1. Set P2D or P3D as the renderer.
2. Set a high pixel density.
3. Use `filter()`.

```
void setup() {
  size(500, 500, P3D);
  pixelDensity(displayDensity());
  rectMode(CENTER);
  noLoop();
  
  background(0);
  noStroke();
  fill(255, 0, 0);
}

void draw() {
  rect(width*0.5, height*0.5, width*0.5, height*0.5);
  filter(GRAY);
}
```

## Your Environment

 

* Processing version: Processing 4.0b6 (x64)
* Operating System and OS version: macOS 12.2.1
* Other information: Apple M1 chip

codeanticode:
This fixes https://github.com/processing/processing4/issues/429 but has not been tested extensively

### DIFF
--- a/core/src/processing/opengl/PGraphicsOpenGL.java
+++ b/core/src/processing/opengl/PGraphicsOpenGL.java
@@ -6153,25 +6153,25 @@ public class PGraphicsOpenGL extends PGraphics {
     int scrX0, scrX1;
     int scrY0, scrY1;
     if (invX) {
-      scrX0 = (dx + dw) / pixelDensity;
-      scrX1 = dx / pixelDensity;
+      scrX0 = (dx + dw) / src.pixelDensity;
+      scrX1 = dx / src.pixelDensity;
     } else {
-      scrX0 = dx / pixelDensity;
-      scrX1 = (dx + dw) / pixelDensity;
+      scrX0 = dx / src.pixelDensity;
+      scrX1 = (dx + dw) / src.pixelDensity;
     }
 
     int texX0 = sx;
     int texX1 = sx + sw;
     int texY0, texY1;
     if (invY) {
-      scrY0 = height - (dy + dh) / pixelDensity;
-      scrY1 = height - dy / pixelDensity;
+      scrY0 = height - (dy + dh) / src.pixelDensity;
+      scrY1 = height - dy / src.pixelDensity;
       texY0 = tex.height - (sy + sh);
       texY1 = tex.height - sy;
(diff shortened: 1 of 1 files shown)


benfry:
Just to be clear, the behavior should be that there are 2x the pixels in either direction, so was the problem just a scaling issue? (i.e. we don't want to just take low-res buffers and scale them larger… `pixelDensity(2)` means all pixel operations live in a 2x space (`pixelWidth` and `pixelHeight` instead of `width` and `height`)

codeanticode:
I think the coordinate scaling is the correct solution because the issue is caused by the texture backing the drawing surface having double the size being rendered on window coordinates...

But, this rescaling should be done only if we are applying copy to a backing surface texture, not a regular image. This is, when copy(PImage src...) is called from copy(int sx, int sy...)

codeanticode:
@benfry  just realized that any PImage contains its own pixelDensity so maybe the last commit fixed properly the issue? We would need to test more, because this change may have unintended consequences :-)

benfry:
Yeah, this seems more correct. Let me know what you find out with testing, and I can merge if you think it's stable and working. Or if you have high confidence now, I can merge and hopefully someone using the source might catch issues before any release.

codeanticode:
I will do some testing with the filter function, hopefully this fixes the issue without any side effects.

benfry:
Thanks!

PR #435, opened 2022-03-04
Renamed parameters controlling GL buffer settings

codeanticode:
Consider this PR to make the name of these parameters more consistent. We might need to play with them until a optimal configuration is found (moment in which we can set them as protected) or perhaps they have to left public for advanced users to tweak these settings depending on the platform or usage mode. Not sure at this point if there is one configuration that fits all situations in terms of performance.

### DIFF
--- a/core/src/processing/opengl/PGL.java
+++ b/core/src/processing/opengl/PGL.java
@@ -77,16 +77,6 @@ public abstract class PGL {
   protected static boolean USE_DIRECT_BUFFERS = true;
   protected static int MIN_DIRECT_BUFFER_SIZE = 1;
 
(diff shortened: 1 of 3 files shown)


benfry:
Ok, though all caps is for unchanging constants (not just static variables). So if these are things that folks might want to modify, they should be camelCase. 🐪 🐫

codeanticode:
I already made them all camelCase:

bufferStreamingImmediate
bufferStreamingRetained
bufferUsageRetained
bufferUsageImmediate
bufferMapAccess

So this PR is ready to merge into beta8 :-)

benfry:
Great. all set.

Issue #279, opened 2021-08-14
IDE auto-formatting confused by method references (40b1)


MichelBillaud:
## Description

The IDE auto-formatting is confused by method references.  Eg

~~~java
  Predicate<Foo> p = Foo::check;
~~~

turns into 

~~~java
::  Predicate<Foo> p = Foocheck;
~~~

## Steps to reproduce

1. Start a new sketch with

~~~java
import java.util.function.Predicate;

class Foo {
  boolean check() {
    return true;
  };
}

void setup() {
  Predicate<Foo> p = Foo::check;
}
~~~

2. `Ctrl+T`

## Environment

* Processing version: 40b1
* Operating System and OS version: Linux Debian 10.10
* Other information:

benfry:
@sampottinger any chance the new preproc might also be able to give us a better formatter?

PR #436, opened 2022-03-04
Fix IDE auto-formatting confused by method references (40b1) #279

TN8001:
This text is an automatic translation.

The build at hand is working correctly with this change.

I am not familiar with java and do not know how to run tests.
I have not been able to verify the test.


benfry:
Thanks; incorporating this for 4.0 beta 8. Fingers crossed!

Issue #453, opened 2022-03-21
README.md in build gives intructions for JDK 11 while project requires JDK 17


toshan-luktuke:
## Description

The current README in build that gives instructions for building Processing 4 all links to the JDK 11 release of Open-JDK, whereas the project currently uses JDK 17.
## Expected Behavior

README should have instructions to install JDK 17.

## Your Environment

 

* Processing version: Processing 4.0 beta 7
* Operating System and OS version: Windows 10
* Other information:

## Possible Causes / Solutions

Change the README in build so that it suggests the required version

benfry:
Was waiting until we were sure about the move to 11. With 17 now available via `apt` on Raspberry Pi, we're safe to make the switch.

PR #454, opened 2022-03-21
Change build README JDK versions from 11 to 17 as per #453

toshan-luktuke:
I've made changes to the JDK version the README talks about from 11 to 17 at all points except the one under the Code Signing subheading. As I'm not sure if its appropriate to change that or not..
Also updated the `Instructions for installing Java` under the `Getting Java and Ant` subheading to point to the new https://adoptium.net/installation.html# link
Resolves issue #453

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -1,18 +1,17 @@
-How to Build Processing
-=======================
+# How to Build Processing
 
 The short version:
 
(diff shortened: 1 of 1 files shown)


benfry:
Can't incorporate this with all the unrelated formatting changes.

PR #459, opened 2022-03-28
Fix 442: Error with tweak mode when using underscores in numbers

### LINKED ISSUE #442: Error with tweak mode when using underscores in numbers
## Description
An error when running tweak mode.

## Expected Behavior
I expect code that runs in normal mode, to not raise a syntax error in tweak mode.

## Current Behavior
When running tweak mode, under a specific set of circumstances, it can raise an error when running normal does not.
It seems that when using underscore syntax for numbers, `1_000` tweak mode does not probably convert this to a tweak integer. 

## Steps to Reproduce
I have found this code to consistently throw the error
```
void setup(){
  size(100, 100);
}

void draw(){
 int x = 1_000;

}
```
the error raised is `Syntax Error - Missing name or ; near ‘ int x = tweakmode_int[0]_000’?`
This error does not happen if there is no `size()` call, other calls or assignments in setup do not raise the error. 
The code does not need to assign 1_000, it can try and pass it to a function or output it.
Underscores in other places in the number also raise the error.

## Your Environment

 

* Processing version:  4.0b7
* Operating System and OS version: Windows 11 Home, 21H2

arnoudvanderleer:
Fix https://github.com/processing/processing4/issues/442: In tweak mode, (hexa)decimal numbers with underscores will be parsed correctly, and have their underscores removed.

### DIFF
--- a/java/src/processing/mode/java/tweak/SketchParser.java
+++ b/java/src/processing/mode/java/tweak/SketchParser.java
@@ -115,7 +115,7 @@ public class SketchParser {
     // for every number found:
     // save its type (int/float), name, value and position in code.
 
-    Pattern p = Pattern.compile("[\\[\\{<>(),\\t\\s\\+\\-\\/\\*^%!|&=?:~]\\d+\\.?\\d*");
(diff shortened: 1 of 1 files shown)


arnoudvanderleer:
The current code does not give an error when a number contains an underscore, but it removes underscores from numbers when tweak mode is started.
I can add (potentially much) code to re-add underscores in tweak mode, but I am not sure whether someone would be interested in this.

benfry:
Thanks!

Issue #450, opened 2022-03-17
Add some information to the readme file


arnoudvanderleer:
The `/README.md` file currently is mainly aimed at people who already know what Processing is and roughly how it works, and is mainly a changelog. It would benefit new developers to have a bit of an explanation of what Processing is and how it works and maybe a small guide to the layout of the repository and/or links to other materials like `processing.org`, `CONTRIBUTING.md` etc.

If I have time, I will make a draft for this in the near future, if you want me to.

benfry:
That's intentional; we have a massive site where we've put a ton of work into getting folks up and running. This is just the repository for the source code. It's not like a library that you might check out the code to include it with your project. 

To avoid confusion, I've added some notes to clarify this in https://github.com/processing/processing4/commit/a4c1d94646f73b4cea5483324bb56a8bb7bcdf1e

PR #460, opened 2022-03-29
Update changelog and readme files

ljystu:
As mentioned in [issue(450)](https://github.com/processing/processing4/issues/450), we made some changes linking relevant websites and files about processing in the `README.md` file. 
We also moved processing changelogs to `CHANGELOG.md`.


benfry:
Thanks, though the current setup is by design. We don't want this repository to be the starting point for people using Processing, and have put a lot of effort into other resources that are better. I also don't want to be in a position of duplicating the same information in two places, because that's even more maintenance for me when there are changes.

Issue #403, opened 2022-02-11
Some keys can't be entered on the On-Screen Keyboard


TN8001:
This was translated by a machine.
## Description
Starting with Processing 4.0 Beta 2, some keys can't be entered with the on-screen keyboard.

## Steps to Reproduce
Open On-Screen Keyboard (`Ctrl` + `Windows` + `O`)

I can enter it.  
alphabet symbol numeric `Tab` `Enter` `Space` etc.

I can't enter it.  
`BS` `Del` `arrow keys` `Home` `PgUp` etc.

## Your Environment
* Processing version: 4.0b2, 4.0b5
* Operating System and OS version: Windows10

## Possible Causes / Solutions
Since `getModifiers()` was changed to `getModifiersEx()` in #67, the `BUTTON1_DOWN_MASK` flag is included.

I don't think there is a combination with mouse buttons for now or probably in the future, so remove the mouse button mask flag?

```java
int modifiers = evt.getModifiersEx();
for (int i = 0; i < MouseInfo.getNumberOfButtons(); i++) {
    modifiers &= ~InputEvent.getMaskForButton(i + 1);
}
```

PR #461, opened 2022-03-31
Fix Some keys can't be entered on the On-Screen Keyboard #403

TN8001:
Fix Some keys can't be entered on the On-Screen Keyboard #403

Perhaps it would be more correct to check the number of buttons as in Issue.
But for most users, this would be a useless process.

Therefore, we decided to remove up to `BUTTON3_DOWN_MASK`, which has a constant.
This is simple and high performance.


benfry:
Thanks; we'll give it a shot!

Issue #447, opened 2022-03-12
Text caret position shifts when typing Japanese


TN8001:
This was translated by a machine.
Sorry for the delay in reporting this, as it stayed in beta1 due to #403 issues.

## Description
Japanese characters are usually twice as wide as alphabetic characters. (Even in a monospace font!!!)
The caret position seems to be confused because all characters are determined to be the same width.

Many fonts are twice as wide (two alphabetic characters for one Japanese character), but some are not.
![fontsize](https://user-images.githubusercontent.com/28250188/157993889-d2e6e87b-dc45-4b1a-9a3d-47fbf1835ca9.png)

## Expected Behavior
There was no problem with beta4.

https://user-images.githubusercontent.com/28250188/157993967-38ff2b17-2e1b-4a93-b165-15c722c6d032.mp4

## Current Behavior
The video is beta5, but the same applies to beta6 and beta7.

https://user-images.githubusercontent.com/28250188/157993978-e7d85d02-1ff4-421a-8dca-c5e05105db81.mp4

## Your Environment
* Processing version: 4.0b5 or later
* Operating System and OS version: Windows10

![settings](https://user-images.githubusercontent.com/28250188/157994089-2b73fd13-2adc-476f-9033-3d807318db07.png)

benfry:
This is likely due to the fixes for https://github.com/processing/processing4/issues/194, https://github.com/processing/processing4/issues/226, https://github.com/processing/processing4/issues/342 which were major problems that prevented people from working at all. 

Do I understand correctly that Japanese works as long as you use a truly monospace font? Does it work correctly if you use the default monospaced font (meaning that you don't change anything in Preferences)? I'm just trying to understand the severity of the issue.

TN8001:
> Do I understand correctly that Japanese works as long as you use a truly monospace font? 

There is no true monospace font (one alphabetic character for one Japanese character) in Japanese as far as I know.
This is because kanji have so many strokes that they cannot be expressed or read in the same width.
This is an exaggerated example (not a commonly used kanji), but kanji require a larger width than the alphabet.
[Taito (kanji) - Wikipedia](https://en.wikipedia.org/wiki/Taito_(kanji))

> Does it work correctly if you use the default monospaced font (meaning that you don't change anything in Preferences)? 

To restore the default settings, do I just delete "C:\Users\[UserName]\AppData\Roaming\Processing"?

Language: Japanese
Font: "Souce Code Pro" (not sure why this was chosen)

[...]

PR #462, opened 2022-03-31
Fix Text caret position shifts when typing Japanese #447

TN8001:
Fix Text caret position shifts when typing Japanese #447

Since the content has changed so much, I left the original implementation in the comments, what do you think?


benfry:
Yes, thanks for leaving it in there for reference—as you've probably seen, I'll often do that for something like this that may need to be reverted/reviewed again shortly after. 

Appreciate the help; I'll get this in for beta 8!

Issue #469, opened 2022-04-20
preprocess() should return PreprocessorResult object


codeanticode:
Currently, the preprocess() method in JavaBuild [returns](https://github.com/processing/processing4/blob/master/java/src/processing/mode/java/JavaBuild.java#L382) a string value containing the class name of the sketch, that is queried from the PreprocessorResult object created inside the method. However, it might be better it if preprocess just returns the PreprocessorResult object, which contains additional useful information such as the sketch's renderer.

benfry:
#470 now incorporated for 4.0b8.

PR #470, opened 2022-04-20
preprocess() method in JavaBuild returns PreprocessorResult object

codeanticode:


codeanticode:
This is to fix #469

sampottinger (approved):
Thanks @codeanticode! I think this would work. Let me branch from your branch and try to get the full screen situation sorted.

Issue #471, opened 2022-04-20
fullScreen() does not work for second display with P2D/P3D in certain screen alignments 


HectorCarral:
## Description

Using `fullScreen()` with P2D or P3D fails to run the sketch on a secondary screen, despite specifying the display number and also regardless of the Processing settings. It always runs on the main screen.

## Expected Behavior

The sketch is run fullscreen on the specified screen, regardless of the renderer used.

## Current Behavior

The sketch always runs fullscreen on the main screen if P2D or P3D is specified.

For example, using a laptop with a secondary screen connected, `fullScreen(1)` correctly runs the sketch fullscreen on screen 1 (main one) while `fullScreen(2)` correctly runs the sketch fullscreen on screen 2 (secondary one). However, when specifying another renderer (`fullScreen(P2D, 2)` or `fullScreen(P3D, 2)`), the sketch runs on screen 1 (main one) while the following message is presented: `The sketch has been resized from 100✕100 to 1920✕1080 by the window manager.` (where 1920✕1080 is indeed the correct resolution of screen 2, different from the resolution of screen 1). This is independently of the setting of default screen specified on the Processing IDE settings. The workaround to run the sketch fullscreen on the external screen is to go into the OS settings and set that as the main screen. Please note that this is on macOS, and I have been unable to test it on other operating systems.

## Steps to Reproduce

1. Have a secondary screen connected.
2. Use `fullScreen()` while specifying P2D/P3D and the number of the secondary screen: 

```
void setup() {
  fullScreen(P3D, 2);
}
```

## Your Environment

 

* Processing version: Processing 4.0b7 (x64)
* Operating System and OS version: macOS 12.3.1
* Other information: Apple M1 chip

sampottinger:
@benfry this is working for me on macOS 11.6 Intel on current master (which includes #474).

```
int x = 0;

void setup() {
  fullScreen(P3D, 2);
  background(0);
  noStroke();
  fill(102);
}

void draw() {
  rect(x, height*0.2, 1, height*0.6); 
  x = x + 2;
}
```

and 

```
int x = 0;

void setup() {
  fullScreen(P3D, 1);
  background(0);
  noStroke();
  fill(102);
}

void draw() {
  rect(x, height*0.2, 1, height*0.6); 
  x = x + 2;
}
```

I do get a resize message from window manager (`The sketch has been resized from 100✕100 to 2560✕1440 by the window manager.`) but it is still functional.

LMK how you want to proceed.

[...]

PR #474, opened 2022-04-21
Find sketch renderer in full screen preproc.

sampottinger:
CC @codeanticode - this should complement #470.


benfry:
Does this also resolve #471?

sampottinger:
Let me check on that @benfry

PR #481, opened 2022-04-28
Update PDE_es.properties

HectorCarral:
Update Spanish translations for new text about managing libraries, tools and modes.


benfry:
Thanks again!

PR #483, opened 2022-04-28
Update PDE_de.properties

hkiel:
Update German translation for new texts.


benfry:
Excellent, thank you!

PR #493, opened 2022-05-28
Fix parsing of java error messages containing ":"

### LINKED ISSUE #492: Switch error only shows last line if caused by other program error
## Description

The following code displays the error `' on 'switch' has been suppressed, which is perhaps related to this problem` without displaying the text of the actual error that caused the problem.

```java
enum Option {
  CAT, DOG;
}

void setup() {
  Option option = Option.CAT;
  String noise;

  switch (option) {
  case CAT:
    noise = "miaow";
    break;
  case DOG:
    noise = "woof";
    break;
  }
  println(noise);
}
```

The full and correct error message is `The local variable noise may not have been initialised. Note that a problem regarding missing 'default' on 'switch' has been suppressed, which is perhaps related to this problem`. Without this context, the 'on switch has been suppressed' message will be confusing, especially to newer programmers.

The full message is temporarily displayed in the status line above the console when typing the line that first attempts to use 'noise' (`println(noise);` above), but not on running the program.

## Expected Behavior

I would expect the full message to be displayed in the console on running. Or perhaps more helpfully just the first sentence `The local variable noise may not have been initialised.` since the switch error is a bit misleading.

## Your Environment

* Processing version: 4 beta 8
* Operating System and OS version: MacOS 11.6.5 (Bug Sur)

dzaima:
Fixes #492, allowing a message like `The local variable noise may not have been initialized. Note that a problem regarding missing 'default:' on 'switch' has been suppressed, which is perhaps related to this problem` to be printed correctly.

### DIFF
--- a/java/src/processing/mode/java/Compiler.java
+++ b/java/src/processing/mode/java/Compiler.java
@@ -133,7 +133,7 @@ public class Compiler {
 
         // get first line, which contains file name, line number,
         // and at least the first line of the error message
-        String errorFormat = "([\\w\\d_]+.java):(\\d+):\\s*(.*):\\s*(.*)\\s*";
+        String errorFormat = "([\\w\\d_]+\\.java):(\\d+):\\s*([^:]*):\\s*(.*)\\s*";
         String[] pieces = PApplet.match(line, errorFormat);
         //PApplet.println(pieces);
 


benfry:
Thanks!

Issue #485, opened 2022-05-09
NoClassDefFoundError: com/formdev/flatlaf/FlatLaf


rupeshkumar22:
## Description
First I downloaded the processing4 master zip file directly and executed git init. After building with `cd build`, I imported the project to eclipse and set the base class in run configurations. When I run it along with Android Mode (via VM arguments) or without Android Mode, I see this error.

## Expected Behavior
I remember no such issue faced on the earlier version of Processsing4.

## Current Behavior
Sharing a snapshot of the issue on running the app -
![Capture](https://user-images.githubusercontent.com/46577873/167476767-a937581d-a828-4510-80a7-ce95cb569aa6.PNG)

## Your Environment
java -version
openjdk version "17.0.1" 2021-10-19
**OS-** Windows 10

## Possible Causes / Solutions
I can see the jar at the `/app/lib/flatlaf.jar` location. Seems that some classpath variables need to be declared. Or maybe I need to use java 17.0.2.

benfry:
If you're building with Eclipse, you'll have to add that .jar file. Eclipse isn't supported; just use `ant run` from inside the `build` folder for the supported way to build/run the software from the repo.

rupeshkumar22:
@benfry adding jar in build path resolved it. thanks! I need to use eclipse for adding Android Mode's VM Arguments. `ant run` also works well.

PR #498, opened 2022-06-14
Resolving ant run

rupeshkumar22:
The splash screen does not move forward towards the base class. The issue arises at runtime. Including flatlaf.jar as a classpath worked for me.

Steps to reproduce and resolve-

- clone repo

- `ant run` inside the build folder

- The splash screen does not move forward.

- close the splash window and open task manager on windows.

- Right Click on OpenJDK Platform Binary -> End Task(otherwise ant shows an error due to some resource locking by java binary)

- add <cp>lib/flatlaf.jar</cp> under classpath tag of `windows/config.xml`

- execute `ant run` again. (worked for me).

Linking a related issue #485

### DIFF
--- a/build/windows/config.xml
+++ b/build/windows/config.xml
@@ -23,6 +23,7 @@
     <cp>lib/jna-platform.jar</cp>
     <cp>lib/ant.jar</cp>
     <cp>lib/ant-launcher.jar</cp>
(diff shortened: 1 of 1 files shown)


benfry:
Thanks for the catch!

PR #504, opened 2022-06-21
Fix #444 by removing old parsing.

### LINKED ISSUE #444: IDE treats {} brackets as code when they're inside a String, char or comment.
When putting a { or a } on a comment, String, or char, the IDE complains of a missing bracket.
The program compiles fine when this happens, but it prevents most other errors from showing up, making it annoying to debug other code, and it's very hard to spot when the project is large.

image
image
If it helps, this is Processing 4.0b6 and I'm on Windows 10.

sampottinger:
It looks like there's still some left over non-ANTLR parsing to check for missing braces from before the new preproc. The new preproc will catch the error and testing makes it look like this old regex check is not triggering anyway outside of comments (which is an issue reported in #444). I may be missing something here though @benfry... LMK what you think. Thanks!

### DIFF
--- a/java/src/processing/mode/java/ErrorChecker.java
+++ b/java/src/processing/mode/java/ErrorChecker.java
@@ -101,11 +101,6 @@ class ErrorChecker {
       problems.addAll(curlyQuoteProblems);
     }
 
-    if (problems.isEmpty()) { // Check for missing braces
-      List<JavaProblem> missingBraceProblems = checkForMissingBraces(ps);
-      problems.addAll(missingBraceProblems);
-    }
-
     if (problems.isEmpty()) {
       AtomicReference<ClassPath> searchClassPath =
         new AtomicReference<>(null);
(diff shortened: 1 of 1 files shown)


benfry:
IIRC this is in place because 1) it can produce exceptionally confusing errors (especially for beginners), and 2) many moons ago, it could put the compiler into a bad state and cause a crash.

I assume the latter is resolved these days, but are we in good shape for the former?

sampottinger:
Thanks @benfry! For sure, the later is fixed. For the former... I'm not quite sure how it would get triggered because, in the case of a missing `{` or `}`, it would fail and finish execution in the preprocessing stage before it hits the old parsing logic anyway. So, right now, it can only be activated if the code is valid (it can only report an error if an error doesn't exist). Two options:

 - Maybe we could modify this to only check for the curly in the case of preproc failure? That being said, the logic does fail for the comments so there's still a chance that the old SourceUtils check could highlight "correct" code and miss the actual error.
 - We find a way for the preproc just to provide a more beginner friendly error message without SourceUtils.

What do you think? I might be more inclined towards the second option just to avoid other failure cases for the SourceUtils.

benfry:
Ok, let's give it a shot for the next beta. Thanks!

PR #505, opened 2022-06-23
Fixed incorrect default constructor, PVector.java

EstiaanJ:
Either the comment above the default constructor is incorrect, or the constructor itself was incorrect and did not initialize the values as per the comment.
I fixed it so that it initializes the x,y and z components to zero.


benfry:
The comment is correct. It's not necessary to explicitly initialize variables to zero in Java.

PR #513, opened 2022-07-20
Add Java arguments to enable Anti-Aliasing in bash script.

QazCetelic:
## Without AA
image
## With AA enabled
image

### DIFF
--- a/build/linux/processing
+++ b/build/linux/processing
@@ -107,5 +107,5 @@ then
 else
   # Start Processing in the same directory as this script
   cd "$APPDIR"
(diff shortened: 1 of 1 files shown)


benfry:
I'm already seeing anti-aliased fonts with default Ubuntu 20.04 and 22.04 installs (the supported platforms for Linux). 

What OS version, window manager, etc are you using that this is required?

QazCetelic:
Operating System: Fedora Linux 36
KDE Plasma Version: 5.25.3
KDE Frameworks Version: 5.96.0
Qt Version: 5.15.3
Kernel Version: 5.18.11-200.fc36.x86_64 (64-bit)
Graphics Platform: X11

benfry:
That was my guess; looks like KDE remains a place where this is required somehow. 

Merged in for beta 9; fingers crossed it doesn't break anything else! Thanks for the PR.

PR #534, opened 2022-08-12
java/preproc: Use class lookup to get the sketchname instead of hardcoding it

Mimoja:
So far the sketchname was passed to the PApplet hardcoded as string in the generated
Java. While this approach is absolutely fine it is duplicate information.
For those who directly edit on the java lever (e.g. to get proper IDE support)
this means two points of edit when changing the sketch name.

Outside of the processing IDE there is no understanding of a sketch name only
a project name with a main class.

We therefore propose that the main class should be used as only source of truth
and the PApplet sketch name should be derived from it.
This also allows to extend the main class for multiple binaries from the same source.

Signed-off-by: Mimoja <git@mimoja.de>

### DIFF
--- a/java/src/processing/mode/java/preproc/PdeParseTreeListener.java
+++ b/java/src/processing/mode/java/preproc/PdeParseTreeListener.java
@@ -1225,9 +1225,8 @@ public class PdeParseTreeListener extends ProcessingBaseListener {
 
     footerWriter.addEmptyLine();
     footerWriter.addCodeLine(indent1 + "static public void main(String[] passedArgs) {");
-    footerWriter.addCode(indent2 +   "String[] appletArgs = new String[] { ");
-
-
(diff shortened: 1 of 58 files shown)


Mimoja:
Instead of calling getName() the call to getSimpleName() might also be worth looking, but in the usual and the aforementioned unusual case it is not required

clankill3r:
You changed:

`argsJoiner.add("\"" + sketchName + "\"");`

Into:

`argsJoiner.add("sketchName");`

benfry:
Appreciate the attempt and thanks for the contribution. Unfortunately, this seems too fragile a fix (how does it interact with class permissions? other JDK versions? other platforms? methods for invoking? threads?) for something so fundamental as starting up the sketch, which is not broken at all at the moment. 

(It's also clear the code wasn't tested heavily because of the typo noted above.)

If you don't want to use the quoted sketch name in your own projects, you can change that line to `PApplet.run(SketchName.class)` so that whenever you refactor `SketchName` to something else, it will automatically update in the code.

PR #542, opened 2022-08-24
Updated popStyle() webBrief

### LINKED ISSUE #271: popStyle() reference error
> fyi only - there appears to be an error in the new documentation at
> https://processing.org/reference  --  the definition for popStyle appears to contain an error (pop instead of push?)
> 
> popStyle()
> Saves the current style settings and popStyle() restores the prior settings

SableRaf:
Fixes https://github.com/processing/processing-website/issues/271


benfry:
Thanks!

PR #585, opened 2022-10-29
Update Ukrainian language strings

rzats:
Fixes two strings modified since the last internationalization update ("add X" → "manage X").


benfry:
Thank you!!

PR #597, opened 2022-11-12
Static with class definition fix.

### LINKED ISSUE #579: function/variable "does not exist" errors reported if trying to define a class without setup/draw
## Description

I was starting a new project and had a mostly empty file as the main PDE, with the exception of an import for the video library. I created a new file to write a class, and all references to Processing functions & variables were red underlined as errors. Error messages were, for example, "mouseX does not exist", despite syntax highlighting picking them up correctly. I was able to make this go away by adding empty setup() and draw() definitions in the main PDE file.

## Expected Behavior

A descriptive error suggesting to define setup() and draw() in the appropriate file before defining a class that depends on Processing variables and functions. 

## Current Behavior

App says library variables and function names like ellipse() and mouseX do not exist.

## Steps to Reproduce

1. Create and save a new sketch
2. Add a second tab with a name of your choice, in my case it was Button
3. Define a class in the new tab, referencing some of the Processing functions, but do not add code to the main file

Here's the code used in my second tab that reproduces the error:
```
class Button {
  
  int x, y, radius;
  
  public Button (int x, int y, int radius) {
    this.x = x;
    this.y = y;
    this.radius = radius;
  }
  
  boolean over() {
    return dist(mouseX, mouseY, this.x, this.y) < this.radius;
  }
  
  void draw() {
    ellipse(this.x, this.y, this.radius * 2, this.radius * 2);
  }
  
}
```
mouseX, mouseY, and ellipse() will show as errors, hovering/clicking for more detail says "does not exist"

## Your Environment

 

* Processing version: 4.0.1
* Operating System and OS version: macOS 12.4 / Apple M1
* Other information:

Based on what I remember about how Processing parses the main sketch file depending on whether or not setup() or draw() are defined, I think it makes sense that it wouldn't be possible to define a (nested) class without formally defining setup()/draw(). But, initially I thought this was some kind of parsing error and restarted the app to see if it would go away. I'm *not* suggesting there's a mode of use that justifies defining a class without setup()/draw(), just that a more descriptive error message would be useful when this happens.

sampottinger:
Closes #579:

It looks like Processing 3 was a bit more eager to try to put the user in static mode. I’m not 100% sure but I think this might have accidentally caused some undesired wrapping previously. Regardless, I replicate the 3x behavior while still supporting Java mode by:

 - Try matching static mode before Java mode.
 - If there is ambiguity, check for main in the root object provided by the user.
 - Check if the user is trying to import processing core manually and avoid double import of automatic includes.

This PR adds tests both from #579 and also a new test where the mode is ambiguous but root object includes a main method.

### DIFF
--- a/java/src/processing/mode/java/preproc/PdeParseTreeListener.java
+++ b/java/src/processing/mode/java/preproc/PdeParseTreeListener.java
@@ -86,6 +86,7 @@ public class PdeParseTreeListener extends ProcessingBaseListener {
   private boolean sizeIsFullscreen = false;
   private boolean noSmoothRequiresRewrite = false;
   private boolean smoothRequiresRewrite = false;
+  private boolean userImportingManually = false;
   private RewriteResult headerResult;
   private RewriteResult footerResult;
 
@@ -443,6 +444,10 @@ public class PdeParseTreeListener extends ProcessingBaseListener {
 
     foundImports.add(ImportStatement.parse(importStringNoSemi));
 
+    if (importStringNoSemi.startsWith("processing.core.")) {
+      userImportingManually = true;
+    }
+
     delete(ctx.start, ctx.stop);
   }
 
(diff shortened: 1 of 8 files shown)


sampottinger:
I think this is ready for you @benfry!

sampottinger:
… Just to be sure, quick checked this on a fresh linux machine and it worked!

Static Test

benfry:
Thanks!

Issue #607, opened 2022-11-27
Space oddities in `.java` code generated by preprocessor


benfry:
A few oddities in the generated code from the preprocessor that should be easy to fix. @sampottinger You mentioned you might look at preproc things today, I'm hoping these might be quick?

* There's an extra space character added before functions (perhaps only functions where `public` has been added? not sure).
* Several extra blank lines of text appear toward the top of the file (after imports and before `public class` iirc)
* None of the rest of the code is indented at all: i.e. once the sketch is wrapped in `public class Blah extends PApplet`, the inner lines have no indent.

I'd also like to rewrite the “exported at” text, but can take care of that.

benfry:
Closing as resolved with #609.

Issue #519, opened 2022-07-30
`warnMixModes` temporarily disabled


benfry:
The `warnMixedModes` section of the Processing.g4 grammar was breaking autocomplete. 

(See https://github.com/processing/processing4/issues/491 for that issue)

The commit that caused the breakage is https://github.com/processing/processing4/commit/0f9b290398aac86ef56fe3c4e9c3ff88fdd53466

The temporary workaround is in https://github.com/processing/processing4/commit/caf1ba90b63d7c6fbcdfbe0586cf848cb353d1dc

This will affect 4.0 beta 9 unless we happen to get a fix before then.

benfry:
@sampottinger This was a nasty one that I meant to ask you about…

sampottinger:
Thanks for catching that @benfry. I should have some time tomorrow. I'll take a look!

benfry:
Awesome, thanks!

PR #609, opened 2022-11-27
Various fixes for preproc

### LINKED ISSUE #602: size(863,863/2) not working as first line of setup() function in a simple sketch
When running one of Dan Shiffman's examples, it failed on the size(863,863/2); line in the setup(). The error said to look at the documentation for the size() function. The only requirement was for the size() function to be in the setup() function as the first line and for it to be in the processing sketch named the same.

However, it appears the size() can't have arithmetic expressions as part of the argument list.

 

## Description

When running one of Dan Shiffman's  mandelbrot example sketch, it failed on the size(863,863/2); statement in the setup() function. The error said to look at the documentation for the size() function. 

The documentation for size():
- size() function must be in the setup() function as the first line 
- and for size() to be in the main processing sketch.

However, it appears the size() can't have arithmetic expressions as part of the argument list.

## Expected Behavior

expected the size(863,863/2); to work or with integer variables as arguments 

## Current Behavior

sketch won't run, error messages says size() can't be used here, changing it to size(863,431); got it to work.

## Steps to Reproduce

1. include a size(863,863/2) as the first statement in the setup() function of a simple sketch
2. 
```java
void setup() {
    size(863,863/2);
}
void draw() {

}
```
3. run program
4. should see error below sketch code

## Your Environment

 

* Processing version: 4.0.1
* Operating System and OS version: MacOS 10.13 Ventura
* Other information: Running on a macmini with Apple M1 chip

workaround is to do the sizing math and keep size(863,431); arguments as simple integer constants. 

provide a better error message that points to the arguments as not being simple integers.

sampottinger:
Addresses some small misc issues in preproc…

 - Closes #602. It will not rewrite if there is a method call or call with variable, passing up to compiler to determine if it is still valid. This is required because of someone overloading size or implementing their own (see #317).
 - Most of #607. See below.

### DIFF
--- a/java/src/processing/mode/java/preproc/PdeParseTreeListener.java
+++ b/java/src/processing/mode/java/preproc/PdeParseTreeListener.java
@@ -693,17 +693,17 @@ public class PdeParseTreeListener extends ProcessingBaseListener {
     if (isSize && argsContext.getChildCount() > 2) {
       thisRequiresRewrite = true;
 
-      sketchWidth = argsContext.getChild(0).getText();
-      boolean invalidWidth = PApplet.parseInt(sketchWidth, -1) == -1;
-      invalidWidth = invalidWidth && !sketchWidth.equals("displayWidth");
-      if (invalidWidth) {
+      boolean widthValid = sizeParamValid(argsContext.getChild(0));
+      if (widthValid) {
+        sketchWidth = argsContext.getChild(0).getText();
+      } else {
         thisRequiresRewrite = false;
       }
 
-      sketchHeight = argsContext.getChild(2).getText();
-      boolean invalidHeight = PApplet.parseInt(sketchHeight, -1) == -1;
-      invalidHeight = invalidHeight && !sketchHeight.equals("displayHeight");
-      if (invalidHeight) {
+      boolean validHeight = sizeParamValid(argsContext.getChild(2));
+      if (validHeight) {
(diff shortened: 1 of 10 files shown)


sampottinger:
@benfry - addressed #602 but looking at #607

sampottinger:
@benfry - I was able to get most of this sorted except for indent. The problem is I have to keep the EDT and Javac in sync with each other when edits are made by preproc so errors and warnings go to the right place when they try to find token locations. However, newlines are not tokens on their own in Antlr. So, doing that indent will take a little more thought to ensure I don’t introduce any new bugs.

I’d recommend merging this without the indent fix and addressing indent in a separate PR.

benfry:
Ok, the indent fix is more about avoiding the extra space before each `public` seen here:

Screen Shot 2022-11-27 at 7 56 13 PM

I'm far less worried about adding indents (especially if ANTLR isn't giving you newlines, I totally get that), but we shouldn't be adding extra spaces in there.

sampottinger:
@benfry - Sorry for the confusion. I think the extra space in front was resolved!

benfry:
Perfect, thanks!

benfry:
And https://github.com/processing/processing4/issues/519 is still open, right? (Totally fine, just wrapping up the release here.)

sampottinger:
Yes #519 is still open. :(

benfry:
No problem, thanks again.

PR #611, opened 2022-11-29
Check for sanitary folder names

### LINKED ISSUE #610: Processing doesn't recognize methods of PApplet when the sketchfolder has a hyphen in it
## Description

When the name of a sketch folder contains a hyphen, Processing doesn't recognize the methods of `PApplet` like `draw()`, `background()`, etc.
The sketch will compile and run without any problems anyway.

## Steps to Reproduce

1. Create a sketch in a folder with a name containing a hyphen
2. A `sketch.properties` file is created pointing to the main file (which can't have a hyphen in it, Processing will refuse to start in that case)
3. Processing won't recognize the methods of `PApplet` anymore (`draw()`, `background()`, etc.)

## Your Environment

 

* Processing version: 4.1.1
* Operating System and OS version: TUXEDO OS 1 (KDE Plasma) / Windows 11
* Other information: This problem exists since 4.1.0

mythicaleinhorn:
This PR fixes #610.

In older versions of Processing, the sketch folder had to have the same name as the sketch itself.
This isn't the case anymore, so folder names have to be checked themselves.

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -1618,7 +1618,7 @@ public class Base {
         Messages.showWarning("You're tricky, but not tricky enough",
           pdeFile.getName() + " is not a valid name for sketch code.\n" +
           "Better to stick to ASCII, no spaces, and make sure\n" +
-          "it doesn't start with a number.", null);
+          "it doesn't start with a number or contains hyphens.", null);
         return null;
       }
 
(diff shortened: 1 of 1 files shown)


clankill3r:
I don't think this is a proper fix, cause a fix would be to make folders with a hyphen work.

benfry:
Thanks, though this doesn't resolve the underlying problem: hyphens should work fine (as folder names, though not class names). #610 also is a problem for spaces in the name or anything else that falls outside the range of acceptable characters, which fundamentally shouldn't be a restriction based on the changes in 4.x.

mythicaleinhorn:
I wasn't aware of that, thank you for clarifying!

PR #622, opened 2022-12-10
Fix static mode visibility injection when annotation present

### LINKED ISSUE #619: cannot use @Override and @Deprecated in static mode
I just updated Processing for version 4.1.1 and now when I run sketch when there is 
`@Override` `@Deprecated` inside `class` Processing don't run, that's a regression ? or there is a goal behind that ?
Before that's work fine withe processing 4.0b7... and all the version before that from 1.5.1

the console return:
```
Cannot find a class or type named @Overridepublic...
```

or 
```
Cannot find a class or type named @Deprecatedpublic...
```

And always bravo for the Processing Team !!!!

sampottinger:
Closes #619 - There is a problem with adding public to members when wrapping in “static” mode processing sketches introduced by removing the leading space in #609. Requires preproc to look for leading annnotations before injecting visibility modifier. Apologies @benfry and @knupel.

### DIFF
--- a/java/src/processing/mode/java/preproc/PdeParseTreeListener.java
+++ b/java/src/processing/mode/java/preproc/PdeParseTreeListener.java
@@ -591,7 +591,7 @@ public class PdeParseTreeListener extends ProcessingBaseListener {
       if (annoationPoint == null) {
         insertBefore(possibleModifiers.getStart(), "public ");
       } else {
-        insertAfter(annoationPoint.getStop(), "public ");
+        insertAfter(annoationPoint.getStop(), " public ");
       }
     }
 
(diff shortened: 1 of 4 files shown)


java/src/processing/mode/java/preproc/PdeParseTreeListener.java
```diff
@@ -591,7 +591,7 @@ public void exitMethodDeclaration(ProcessingParser.MethodDeclarationContext ctx)
       if (annoationPoint == null) {
         insertBefore(possibleModifiers.getStart(), "public ");
       } else {
-        insertAfter(annoationPoint.getStop(), "public ");
+        insertAfter(annoationPoint.getStop(), " public ");
```
sampottinger: @benfry this is the actual fix. Everything else in this PR is just adding additional tests to prevent regressions.

PR #635, opened 2023-01-10
Fix #626 (partial import of processing.core)

### LINKED ISSUE #626: [v4.0.2 and above] “Cannot find a class or type named ‘PApplet’” error
## Description
I’m maintaining an open-source Processing app called [FOSStriangulator](https://github.com/FOSStriangulator/FOSStriangulator) . It used to compile just fine with every Processing version I’ve tried up until Processing 4.0.2. Now, with Processing 4.0.2 and newer versions, it won’t compile, whereas with 4.0.1, it still does. With the newer versions, it’s giving me this error:

`Cannot find a class or type named “PApplet”`

It’s also giving me a range of errors, all seemingly random, such as not being able to find classes like PImage, PrintWriter, IntList, etc.

b7ef6f81331173d25768f9451edd4f4254bd2d7b

I looked at the [4.0.2 release notes 1](https://github.com/processing/processing4/releases/tag/processing-1287-4.0.2) and I can’t see any change that might be causing the issues.

I originally posted about this [on the forum](https://discourse.processing.org/t/processing-4-0-2-and-above-giving-cannot-find-a-class-or-type-named-papplet-error/40110) and I was directed to file a bug report here.

## Expected Behavior
I was expecting the app to run without a problem, just like it does on all the Processing versions I tried before 4.0.2.

## Current Behavior
As described above, running the app on versions 4.0.2 and above gives the `Cannot find a class or type named “PApplet”` error and fails to run.

## Steps to Reproduce
1. Clone the [FOSStriangulator repository](https://github.com/FOSStriangulator/FOSStriangulator).
2. Open the project up in Processing 4.0.2 or above.
3. Run the project.

## Your Environment

 

* Processing version: 4.0.2 and above (I tried every major version)
* Operating System and OS version: macOS 13.1 on ARM; also [independently verified on Linux](https://discourse.processing.org/t/processing-4-0-2-and-above-giving-cannot-find-a-class-or-type-named-papplet-error/40110/2)

sampottinger:
Fixes #626

### DIFF
--- a/java/src/processing/mode/java/preproc/PdeParseTreeListener.java
+++ b/java/src/processing/mode/java/preproc/PdeParseTreeListener.java
@@ -1076,17 +1076,10 @@ public class PdeParseTreeListener extends ProcessingBaseListener {
    */
   protected void writeImports(PrintWriterWithEditGen headerWriter,
         RewriteResultBuilder resultBuilder) {
-
-    if (!userImportingManually) {
-      writeImportList(headerWriter, coreImports, resultBuilder);
-    }
-
+    writeImportList(headerWriter, coreImports, resultBuilder);
     writeImportList(headerWriter, codeFolderImports, resultBuilder);
     writeImportList(headerWriter, foundImports, resultBuilder);
-
-    if (!userImportingManually) {
-      writeImportList(headerWriter, defaultImports, resultBuilder);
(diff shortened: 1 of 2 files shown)


benfry:
Thanks!

PR #639, opened 2023-01-18
Add ability to auto-rename invalid sketch code name

mythicaleinhorn:
When a user tries to open sketch code with an invalid name, they are faced with this dialog:
error_message

From there on they have to manually rename the code and try opening it again.
Since we know the rules and already have a function used to sanitize names,
we can also offer the user to rename it for them:

rename_dialog

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -1453,11 +1453,47 @@ public class Base {
       // Mode that it's using. (In 4.0b8, this became the fall-through case.)
 
       if (!Sketch.isSanitaryName(pdeFile.getName())) {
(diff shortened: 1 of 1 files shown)


benfry:
Probably not a great idea, unfortunately. This doesn't cover situations where a file with the sanitized name already exists, or if we have multiple files that need to be renamed.

In order to hit this situation, the user would have had to manually change the `.pde` file name to something that's not a legitimate name from outside Processing. I'm ok with the fix for that being that they need to fix it the same way they broke it—outside the PDE. It's not ideal, but it's better than a more complicated system for working around a problem not created by/in the PDE itself.

Issue #515, opened 2022-07-26
SVG drawing problem with a specific file


benfry:
Not sure if this is because of the elliptical arc handling or how the path is being closed, but the current (as of 4.0 beta 8) SVG parsing produces this: 

<img width="512" alt="Screen Shot 2022-07-26 at 6 45 53 AM" src="https://user-images.githubusercontent.com/1623101/180988528-2870b8cc-a075-4f2b-8e7a-ab868d9edeb8.png">

for this file:

```
<?xml version="1.0" encoding="utf-8"?>
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" viewBox="0 0 32 32" xml:space="preserve">
<g><path fill="#9B9B9B" d="M16,0A16,16,0,1,0,32,16,16,16,0,0,0,16,0Zm0,25A8,8,0,0,1,16,9h.59l-2.3-2.29a1,1,0,0,1,1.42-1.42l4,4a1,1,0,0,1,.21.33,1,1,0,0,1,0,.76,1,1,0,0,1-.21.33l-4,4a1,1,0,0,1-1.42,0,1,1,0,0,1,0-1.42L16.59,11H16a6,6,0,1,0,6,6,1,1,0,0,1,2,0A8,8,0,0,1,16,25Z"/></g>
</svg>
```

Re-saving the file with Illustrator produces this code:

```
<?xml version="1.0" encoding="utf-8"?>
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" viewBox="0 0 32 32" xml:space="preserve">
[...]

PR #659, opened 2023-01-30
Support chained decimals during SVG Parsing

bsapozhnikov:
@benfry took a stab at the fix discussed for #515 along with a new test that catches the issue.
Currently, when there are 2+ decimals in a row without any whitespace or other separators in-between, we treat the whole thing as a single token. Added some logic to add a separator at the end of each set of decimal digits to handle this case.


benfry:
Thanks! Now merged for Processing 4.2.

PR #673, opened 2023-02-19
fixed #671: Updated processing.data.Table::getString

### LINKED ISSUE #671: Table::getString raises stackOverflowError on double value encounter
## Description

The method Table::getString raises a stackOverflowError when it encounters a double value.

## Expected Behavior

I was expecting to get a double value from the getString method.

## Current Behavior

The getString method fails on returning a double value with a StackOverflowError.

## Steps to Reproduce

1. Create a table with at least 1 column of floating values and an arbitrary amount of rows:
```csv
// data/table.csv
0.001,0.002
0.001,0.002
0.001,0.002
0.001,0.002
```
1. Load and save the table in the setup function:
```processing
// This works perfectly
void setup() {
  Table t = loadTable("data/table.csv", "csv");
  println(t.getString(2, 0));
}
```
3. Then, edit the first column type:
```processing
// This raises a StackOverflowError
void setup() {
  Table t = loadTable("data/table.csv", "csv");
  t.setColumnType(0, Table.DOUBLE);
  println(t.getString(2, 0)); // right there
}
```
4. Run the sketch.

## Your Environment

 

* Processing version: Processing 4.1.3-1, x86_64
* Operating System and OS version: Arch Linux, Garuda Cinnamon 5.6.7, Kernel version: 6.1.12-zen1-1-zen
* Other information:

After further investigation, it seems that the bug originated from the [following statement](https://github.com/processing/processing4/blob/main/core/src/processing/data/Table.java#L3553), when a float is passed to the function call `Double.isNan()` and no else statement is defined:
```processing
// processing.data.Table::getString() - line 3553
    if (Double.isNaN(getFloat(row, column))) {
      return null;
    }
  }
  return String.valueOf(Array.get(columns[column], row));
}
```

I fixed the bug by overriding the method as follow:
```processing
// FixedTable.pde
import processing.data.Table;

class FixedTable extends Table {
  ...

  @Override
  public String getString(int row, int column) {
    checkBounds(row, column);
    if (Table.DOUBLE == this.getColumnType(column)) {
      double value =  this.getDouble(row, column);
      if (Double.isNaN(value)) {
        return null;
      } else {
        return Double.toString(value);
      }
    } else {
      return super.getString(row, column);
    }
  }
}

Elmo54:
Updated the Double.isNaN call value from float to double.

Added an else statement that returns the parsed string value.

### DIFF
--- a/core/src/processing/data/Table.java
+++ b/core/src/processing/data/Table.java
@@ -3550,8 +3550,11 @@ public class Table {
         return null;
       }
     } else if (columnTypes[column] == DOUBLE) {
-      if (Double.isNaN(getFloat(row, column))) {
+      double value = getDouble(row, column);
+      if (Double.isNaN(value)) {
         return null;
+      } else {
+        return Double.toString(value);
       }
     }
     return String.valueOf(Array.get(columns[column], row));


benfry:
Fixed with https://github.com/processing/processing4/commit/011b64b3bfb8929d41e9cbd36eb61252857ceec8, thanks!

PR #678, opened 2023-02-26
LSP feature/declaration support

### LINKED ISSUE #676: Feature: Language Server Protocol, Definition support
## Description
Since there now is [LSP](https://github.com/processing/processing4/pull/564) support for processing, it would be nice if the server would be able to "find declarations"

## Expected Behavior
Finding declarations is supported in the PDE, so it should also be available for the LS

## Current Behavior
Finding declarations isn't supported at the moment

## Your Environment
* Processing version: 4.2

- Add a declarationFinder class to the LSP folder
- Use it to find a declaration using the AST from the preprocessor (Use "InspectMode" as an example)
- Send (found) the declaration back to a LS-client

I'm willing to implement this, so a PR is in the making

Efratror:
Added declaration support by rewriting a part of [InspectMode.java](https://github.com/processing/processing4/blob/ae809133cdeb8566c5b4d252eafd9e8e71814db2/java/src/processing/mode/java/InspectMode.java#L145-L191) and adding this capability to the server.

Please give your honest code-review. As i'm not as experienced in java it was my first time using CompletableFutures and Optionals. There are probably faster ways but i've written them out allot to understand them.

Should resolve #676

### DIFF
--- a/java/src/processing/mode/java/SketchInterval.java
+++ b/java/src/processing/mode/java/SketchInterval.java
@@ -14,9 +14,9 @@ public class SketchInterval {
     this.stopPdeOffset = stopPdeOffset;
   }
 
-  final int tabIndex;
-  final int startTabOffset;
-  final int stopTabOffset;
+  public final int tabIndex;
+  public final int startTabOffset;
+  public final int stopTabOffset;
 
   final int startPdeOffset;
   final int stopPdeOffset;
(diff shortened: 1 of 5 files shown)


benfry:
Thanks! We'll give it a shot for 4.2.1.

Main thing with the code would be following the [style guidelines](https://github.com/processing/processing4/wiki/Style-Guidelines), but the rest looks ok as far as how it integrates with the rest of the source. I don't have time this morning to dig through the code in detail but hopefully having it in use will shake things out a bit.

Great to have more features coming to the LSP implementation!

Efratror:
Thanks for the quick review.

I've seen the style guide and tried to follow them as closely as possible. Clearly i've misunderstood some parts or wasn't consistent enough. Can you please clarify? @benfry

PR #690, opened 2023-03-19
Lsp feature/reference support

### LINKED ISSUE #684: Feature: Language Server Protocol, Reference Support
## Description
Add a feature for supporting references to the LS

## Expected Behavior
References are supported by the PDE (find usage). So it would be nice to have it for the server as well

## Current Behavior
Only finding declarations is supported at the moment

## Your Environment
* Processing version: 4.2
* Other information: Source includes PR #678

* Add static function "searchReference" to the "pdeSymbolFinder" class
* Use it to find all references using the AST,
* Send (found) references back to the LS-client

I'm willing to implement this, please assignee me if you're ok with this idea.

Efratror:
Added reference support by using a part of [ShowUsage.java](https://github.com/processing/processing4/blob/80a5b124f1aa4e08e4d23d1eac97a5b1feb5211a/java/src/processing/mode/java/ShowUsage.java#L172-L181) and added the capability to the server.

Also fixed a code style violation as pointed out in #678.

Should resolve #684

### DIFF
--- a/java/src/processing/mode/java/ASTUtils.java
+++ b/java/src/processing/mode/java/ASTUtils.java
@@ -163,7 +163,9 @@ public class ASTUtils {
   }
 
 
-  protected static List<SimpleName> findAllOccurrences(ASTNode root, String bindingKey) {
+  public static List<SimpleName> findAllOccurrences(ASTNode root,
+    String bindingKey
+  ) {
     List<SimpleName> occurrences = new ArrayList<>();
     root.getRoot().accept(new ASTVisitor() {
       @Override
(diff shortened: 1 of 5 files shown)


sampottinger:
Thanks so much for this @Efratror! We are working through some other stuff but hope to get back soon. CC @SableRaf

Efratror:
No problem, there is more where this came from 😄 Was waiting for your approval for this first so not to overflow the repo with PR's.  In short: take your time i'm not going any were 😉

benfry:
Thank you!

PR #692, opened 2023-03-24
Bugfix for Issue #606

WillRabalais04:
Updates the getStringValue method of the VariableNode class so that it correctly handles multidimensional arrays. The default formatting of multidimensional arrays is to have the size of the first array written in the last set of brackets eg.int[][5]. This changes it so that the returned value has the size of the first array written in the first set of brackets eg.int[5][]. This can parse multidimensional arrays of any size.

PR #713, opened 2023-05-09
Refactor description of arrays in debugger.

### LINKED ISSUE #606: Debugger lists immediate array dimension last
```
void setup(){
  int [][] testArray = new int[3][4];
  println(testArray.length); // set a breakpoint here, if you like
  println(testArray[0].length);
}
```
## Debugger lists immediate array dimension last

when stepping through the above code, the type of `testArray` is reported by the debugger as `int[][3]`, which is easily understood, but a bit confusing.

## Expected Behavior

Since this seems to be the java way of declaring arrays-of-arrays, it would seem more consistent to keep the declared order of dimensions, and call `testArray` an `int[3][]`, EVEN IF it's an array of `int[]` pointers.

## Your Environment

 

* Processing version: 3.4 and 4.0.1
* Operating System and OS version: Linux xyzxyz 6.0.8-arch1-1 ♯1 SMP PREEMPT_DYNAMIC Thu, 10 Nov 2022 21:14:24 +0000 x86_64 GNU/Linux

sampottinger:
Closes #606, builds on #692 from @WillRabalais04.

### DIFF
--- a/java/src/processing/mode/java/debug/VariableNode.java
+++ b/java/src/processing/mode/java/debug/VariableNode.java
@@ -56,7 +56,7 @@ public class VariableNode implements MutableTreeNode {
   public static final int TYPE_VOID = 11;
 
   private static final Pattern ARRAY_REGEX = Pattern.compile(
-    "^(?<prefix>[^\\]]+)(?<unbounded>(\\[\\])*)(?<bounded>(\\[\\d+\\])+).*$"
+    "^(?<prefix>[^\\[]+)(?<unbounded>(\\[\\])*)(?<bounded>(\\[\\d+\\])+)(?<unneeded>[^\\[]*)$"
   );
 
   protected String type;
--- a/java/src/processing/mode/java/debug/VariableNode.java
+++ b/java/src/processing/mode/java/debug/VariableNode.java
@@ -95,6 +95,19 @@ public class VariableNode implements MutableTreeNode {
       } else if (getType() == TYPE_ARRAY) {
         //instance of int[5] (id=998) --> instance of int[5]
         str = value.toString().substring(0, value.toString().lastIndexOf(" "));
+        /*
+        *formats multidimensional array values to have the size of the first array in
+        *the first bracket eg.int[][5]-->int[5][]
+        */
+        // resolves issue #606: https://github.com/processing/processing4/issues/606 
+        if (str.contains("][")) {
+          String brackets = str.substring(str.indexOf('['));
+          int arrayDimensions = 0;
+          String num = brackets.replaceAll("[^\\d]", "");
+          arrayDimensions = (brackets.length() - num.length()) / 2;
+          brackets = "[" + num + "]" + "[]".repeat(arrayDimensions - 1);
+          str = str.substring(0, str.indexOf('[')) + brackets;
+        }
       } else if (getType() == TYPE_STRING) {
         str = ((StringReference) value).value(); // use original string value (without quotes)
       } else {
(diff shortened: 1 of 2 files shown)


sampottinger:
[edited slightly]

Hey @WillRabalais04! Can you do a code review on this and see how my additions feel? It’s built from your #692 (thank you! 👍✨) with a few additions:

 - The function was getting a bit nested so broke out a private method for the array case.
 - I like what you did with the regex but, as I was working with it, I noticed all of your logic could all fit into your regex itself, saving us a branch on if multi-dimensional array.
 - Added tests (use `ant test`)!

Re, regex: We can reverse out on this. Curious to get your opinion. Given the age and multiple strata from multiple generations of contributors (😅) I've found that, for Processing, sometimes consolidating to a documented / tested regex is sometimes a bit easier to maintain when working on some of the string manipulation if the regex isn’t too involved.

sampottinger:
CC @SableRaf - Is it possible to add GSoC folks to the repo if they join our communal pledge not to push direct to main? @WillRabalais04 - how would you feel about that power / responsibility?

SableRaf:
> CC @SableRaf - Is it possible to add GSoC folks to the repo if they join our communal pledge not to push direct to main?

We could also consider adding branch protection to the repository: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches

sampottinger:
Yeah I’m into that. I think we haven’t so far because the list of contributors with upstream access has been small. Requiring pr (with an exceptions list including Ben) to get to main would be ideal either way though.

java/src/processing/mode/java/debug/VariableNode.java
```diff
@@ -380,4 +385,27 @@ public int hashCode() {
...
+   *    int[5] (id=998)" or "instance of int[][5] (id=998)"
+   * @return Human-friendly description like "instance of int[5]" or
+   *    "instance of int[5][]".
+   */
+  private String describeArray(String fullDescription) {
+    Matcher matcher = ARRAY_REGEX.matcher(fullDescription);
+    StringJoiner joiner = new StringJoiner("");
+    if (!matcher.matches()) {
+      return fullDescription;
+    }
```
WillRabalais04, line 404: I moved this line to after the check for whether the pattern matches to prevent creating a new, unused object.

java/test/processing/mode/java/debug/VariableNodeTests.java
```diff
@@ -0,0 +1,101 @@
...
+    Value value = buildMockString("testing");
+    VariableNode node = new VariableNode("test", "java.lang.String", value);
+    Assert.assertEquals(node.getStringValue(), "testing");
+  }
+
+  @Test
+  public void describeSimpleArray() {
+    Value value = buildMockValue("instance of int[5] (id=998)");
+    VariableNode node = new VariableNode("test", "int[]", value);
+    Assert.assertEquals(node.getStringValue(), "instance of int[5]");
```
WillRabalais04, line 51: With the updated regex, one-dimensional arrays no longer match the pattern and the full description is returned.

java/src/processing/mode/java/debug/VariableNode.java
```diff
@@ -52,13 +55,16 @@ public class VariableNode implements MutableTreeNode {
   public static final int TYPE_SHORT = 10;
   public static final int TYPE_VOID = 11;
 
+  private static final Pattern ARRAY_REGEX = Pattern.compile(
+    "^(?<prefix>[^\\[]+)(?<unbounded>(\\[\\])*)(?<bounded>(\\[\\d+\\])+)(?<unneeded>[^\\[]*)$"
+  );
```
WillRabalais04, line 60: Updated regex does not match one-dimensional arrays.

sampottinger:
Superseded by #729

PR #715, opened 2023-05-10
Fix error highlighting in IDE.

### LINKED ISSUE #714: Syntax error highlighting placement / width incorrect.
## Description
Trying back to 4.1, error highlighting doesn't appear to be working (at least on Linux).

## Expected Behavior
Expected a red underline at line 4 or 5.

## Current Behavior
No line shown

## Steps to Reproduce
Use the code provided and wait for error checking:

```
void setup() {
  size(100, 100)
  background(#FFFFFF);
}

void draw() {
}
```

## Your Environment
* Processing version: 4.2
* Operating System and OS version: Pop!_OS 22.04 LTS 
* Other information: x64

I'll check if this is linux specific... possible there's something in theming or attribution.

sampottinger:
Error highlighting is a bit broken (see #714).

**Root cause:** Depending on how the `Problem` is made, the error may be given relative to start of line or start of tab. Changed method names to indicate to users of `Problem`s which one they are working with.

I think this may have been [introduced in the language server support effort as `Problems` may be made without Editor available](https://github.com/processing/processing4/commit/ac062a502dec646dd2f332ea12c32efc01064b1c). The behavior in #714 appears to have started then but there was a delay before it showed up in production given how releases were done.

**Resolution:** I clarified the interface and gave client code ability to determine if they want / need a `Problem `to be placed relative to start of line or start of tab where each are more convenient in different contexts.

### DIFF
--- a/app/src/processing/app/Problem.java
+++ b/app/src/processing/app/Problem.java
@@ -21,16 +21,76 @@
 package processing.app;
 
 
+/**
+ * Structure describing a problem encountered in sketch compilation.
+ */
 public interface Problem {
+
+  /**
+   * Get if the problem is an error that prevented compilation.
+   * 
+   * @return True if an error such that the sketch did not compile and false
+   *    otherwise.
+   */
   public boolean isError();
+
+  /**
+   * Get if the problem is an warning that did not prevent compilation.
+   * 
+   * @return True if a warning and the sketch compiled and false otherwise.
+   */
   public boolean isWarning();
 
+  /**
+   * Get which tab (sketch file) the problem was encountered.
+   * 
+   * @return The index of the tab in which the problem was encountered.
+   */
   public int getTabIndex();
-  public int getLineNumber();  // 0-indexed
+
+  /**
+   * Get at which line the problem was encountered.
+   * 
+   * @return Zero-indexed line number within the tab at getTabIndex in which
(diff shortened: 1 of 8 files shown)


sampottinger:
Hey @benfry! It looks like some ambiguity in `Problem` caused an issue [when we implemented language server support](https://github.com/processing/processing4/commit/ac062a502dec646dd2f332ea12c32efc01064b1c) that broke error highlighting. LMK if this looks good! Sorry I noticed something was off a while ago but I thought it was a theme issue and didn't have time to dig into exactly what was going on until today when I realized it wasn't just my local install. 😅

sampottinger:
CC @SableRaf

sampottinger:
Doing one bit of cleanup this morning but we should prioritize this

sampottinger:
Cool did the clean up! Ready for review... I think it's probably best to have client code indicate which type of start offset it wants when interacting with a problem. I've tested the PDE... checking language server.

sampottinger:
Yay! Confirmed language server working. :D

sampottinger:
I know we don't have an imminent release but this is is getting in the way of a GSoC project. To make sure he can still finish on time / pass, I'm going to merge this to main but I won't make a release. Without this sorted, the error message rewriting isn't visible to the user.

PR #729, opened 2023-06-05
Alternative regex for resolving issue#606 

### LINKED ISSUE #606: Debugger lists immediate array dimension last

WillRabalais04:
Closes #606

### DIFF
--- a/java/src/processing/mode/java/debug/VariableNode.java
+++ b/java/src/processing/mode/java/debug/VariableNode.java
@@ -56,7 +56,7 @@ public class VariableNode implements MutableTreeNode {
   public static final int TYPE_VOID = 11;
 
   private static final Pattern ARRAY_REGEX = Pattern.compile(
-    "^(?<prefix>[^\\]]+)(?<unbounded>(\\[\\])*)(?<bounded>(\\[\\d+\\])+).*$"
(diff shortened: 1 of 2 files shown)


sampottinger (approved):
Thanks for these great builds @WillRabalais04 !

sampottinger:
Merging to help #749 maintain a clean diff.

Issue #751, opened 2023-07-16
Mode-breaking interface change causing regression in `Problem`


benfry:
Looks like a new method added to `Problem` is breaking the API for Modes. Any changes that are made need to allow contributed Modes (p5jsMode, Python, etc) to still run without a recompile so we need a better solution here. Paging @sampottinger?

```
Exception in thread "AWT-EventQueue-0" java.lang.AbstractMethodError: Receiver class processing.mode.p5js.p5jsEditor$2 does not define or inherit an implementation of the resolved method 'abstract int computeTabStopOffset(processing.app.Problem$LineToTabOffsetGetter)' of interface processing.app.Problem.
	at processing.app.ui.Editor.lambda$findProblems$27(Editor.java:2633)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:178)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
	at processing.app.ui.Editor.findProblems(Editor.java:2640)
	at processing.app.syntax.PdeTextAreaPainter.paintErrorLine(PdeTextAreaPainter.java:154)
	at processing.app.syntax.PdeTextAreaPainter.paintLine(PdeTextAreaPainter.java:146)
	at processing.app.syntax.TextAreaPainter.paint(TextAreaPainter.java:243)
	at java.desktop/javax.swing.JComponent.paintToOffscreen(JComponent.java:5311)
	at java.desktop/javax.swing.RepaintManager$PaintManager.paintDoubleBufferedImpl(RepaintManager.java:1657)
	at java.desktop/javax.swing.RepaintManager$PaintManager.paintDoubleBuffered(RepaintManager.java:1632)
[...]

PR #752, opened 2023-07-17
Fix the language server / PDE error highlighting another way

sampottinger:
CC @benfry 

Retry #715. See below for original description.

> Depending on how the Problem is made, the error may be given relative to start of line or start of tab. I think this may have been [introduced in the language server support effort as Problems may be made without Editor available](https://github.com/processing/processing4/commit/ac062a502dec646dd2f332ea12c32efc01064b1c). The behavior in https://github.com/processing/processing4/issues/714 appears to have started then but there was a delay before it showed up in production given how releases were done.

Needs to address #751 in which the `Problem` interface cannot change in any way.

> Looks like a new method added to Problem is breaking the API for Modes. Any changes that are made need to allow contributed Modes (p5jsMode, Python, etc) to still run without a recompile so we need a better solution here. 

CC @benfry - Working on this will keep you posted.

### DIFF
--- a/app/src/processing/app/Problem.java
+++ b/app/src/processing/app/Problem.java
@@ -29,21 +29,6 @@ import java.util.Optional;
 public interface Problem {
 
   /**
-   * Strategy converting line number in tab to character offset from tab start.
-   */
-  public interface LineToTabOffsetGetter {
-
-    /**
-     * Convert a line number to the number of characters past tab start.
-     * 
-     * @param line The line number to convert.
-     * @return The number of characters past tab start where that line starts.
-     */
-    public int get(int line);
-
-  }
-
-  /**
    * Get if the problem is an error that prevented compilation.
    * 
    * @return True if an error such that the sketch did not compile and false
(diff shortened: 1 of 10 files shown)


sampottinger:
Thanks for your patience @benfry and sorry about the modes recompilation impact. I believe this should work while maintaining the `Problem` interface. I left my docstrings in and standardized how everyone was using it (the original issue was that some parts of the code were inconsistent with what the methods were supposed to do / what offset referred to). I have tested:

 - Syntax errors (missing semicolon, braces, etc)
 - Java errors via JDT (missing class)
 - Java errors via javac
 - Runtime errors

CC @WillRabalais04

sampottinger:
Also tried Python mode and it worked. That said, error highlighting wasn't working there. I tried an earlier release of Processing 4 though and it wasn't working in that prior release either so it may be an unrelated issue. LMK if I can help.

benfry:
Great, thanks for the quick turnaround.

PR #757, opened 2023-07-27
migrate from altool to notarytool

### LINKED ISSUE #755: Switch from `altool` to `notarytool` for macOS notarization with Apple
Apple is thinking different again, and we'll need to change the build to use `notarytool` instead of the current `altool`, per this warning during the `ant dist` process:

```
*** Warning: altool has been deprecated for notarization and starting in late 2023 will no longer be supported by the Apple notary service. You should start using notarytool to notarize your software. (-1030)
```

Help!

CyberFlameGO:
I'm not too sure if I've implemented this correctly so please review this but I referred to https://developer.apple.com/documentation/technotes/tn3147-migrating-to-the-latest-notarization-tool in adapting the build

Closes #755

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -902,16 +902,14 @@
     <!-- https://developer.apple.com/account/ -->
 
     <exec executable="/usr/bin/xcrun" dir="macos" failonerror="true">
-      <arg value="altool" />
-      <arg value="--notarize-app" />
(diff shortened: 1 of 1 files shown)


benfry:
Awesome, this is a huge help. Seems to be working with an additional minor edit.

PR #759, opened 2023-07-30
Fix copyBufferFromSource throwing an exception

Clarvel:
If usedBuffers was null and the buffer cache reached its maximum size, copyBufferFromSource would throw a null reference error. This adds a null check to that case to instantiate the usedBuffers list as is done elsewhere in this file.


benfry:
Thanks! We'll give this a shot for the next release.

PR #764, opened 2023-08-05
replace revese(Object list) reverse by reverse(<T> list) and add for longand double

knupel:
Hello, After a long afternoon of test on my algorithm to improve the speed of function `reverse()`. The result it's 90% of fail, the existing algorithm win the race!!! Except for the `reverse(Object list)`where an algorithm with generetic type `reverse(<T> list)` is better like a folk jump from the Earth to the Moon with a single feet !
But I learn something revelant. 
A loop `for` we one operation is better than a loop divide by 2 with two operations in this one :) Why I don't know...I see that on my test with 30_000_000 of elements !
And An other thing I learn with my test. Each time I test few loops and print it, the first loop is always slower. IT's one reason at the first time I think my new algorithm is better :) But I changed the place and add more test... I see the speed of same algoithm increase for unknow reason... I don't understand why that's happen. But its good to know for the next tests, we cannot trust the first `loop for` when there million of elements !!!!

I add funtino revese for long and double... I can remove if necessary ? 
Ot start to work to add double and long function every where in Processing... on my lost time :)

I hope my franglish is not so french !

below the test result for information : 
```java
/**
* reverse function faster than Processing
* may be when I've the possibility I push this one on Processing
*/
import rope.utils.R_Utils.Ru;
import rope.core.Rope;

int num = 30_000_000;

long [] arr = new long[num];
// double [] arr = new double[num];
// float [] arr = new float[num];
// int [] arr = new int[num];
// boolean [] arr = new boolean[num];
// char [] arr = new char[num];

// Float [] arr = new Float[num];
// Integer [] arr = new Integer[num];
// Machin [] arr = new Machin[num];
// PVector [] arr = new PVector[num];
// String [] arr = new String[num];
  

// GOOD CASE

// double
// P5 x1(4270 ms)
// P5 new algo x50(82 ms)
// reverse Rope x120(32.0 ms)
// reverse_copy Rope x20(200 ms)

// long
// P5                  x1(4360 ms)
// P5 new algo         x40(76 ms)
// reverse() Rope      x120(33 ms)
// reverse_copy() Rope x20(227 ms)

// class Float
// P5                x1(3400 ms)
// P5 new algo       x6.5(486 ms)
// reverse Rope      x7(440 ms)
// reverse_copy Rope x6.5(481 ms)

// class Integer
  // P5              x1(3460 ms)
// P5 new algo       x9(490 ms)
// reverse Rope      x8.5(440 ms)
// reverse_copy Rope x8(425 ms)

// PVector
// P5                  x1(3500 ms)
// P5 new algo         x7(500 ms)
// reverse() Rope      x8(390 ms)
// reverse_copy() Rope x5(700 ms)

// class Machin
// P5                 x1(3478 ms)
// P5 new algo        x8(495 ms)
// reverse Rope       x8(453 ms)
// reverse_copy Rope  x8.5(420 ms)

// int
// P5                   x1(34 ms)
// P5 new               0.75(50 ms)
// reverse() Rope       x1.1(28 ms)
// reverse_copy() Rope  x0.25(120 ms)

// float
// P5                   x1(34 ms)
// P5 new               0.7(56 ms)
// reverse() Rope       x1.1(28 ms)
// reverse_copy() Rope  x0.25(123 ms)

// char
// P5                x1(44 ms)
// P5 new algo       x1.1(41 ms)
// reverse Rope      x1.4(29 ms)
// reverse_copy Rope x0.5(85 ms)
 

// BAD CASE
// boolean
// P5                 x1(52)
// P5 new algo        x1(52)
//reverse() Rope      x1.2(43)
//reverse_copy() Rope x0.8(64)

// String
// P5                  x1(390)
// P5 new algo         x0.8(432)
//reverse() Rope       x0.9(416)
//reverse_copy() Rope  x0.8(466)
  
  
void setup() {}

void draw() {
  if(start_is) {
    start_count();
  }
}

void mousePressed() {
  start_is = true;
}

boolean start_is = false;
void start_count() {
  for(int i = 0 ; i < arr.length ; i++) {
    // arr[i] = true;
   arr[i] = i;
    // arr[i] = i;
   // arr[i] = (char)i;
    // arr[i] = ""+i;
    // arr[i] = new PVector(i,i);
    // arr[i] = new Machin();
  }
  
  float start = 0;   

  start = millis();
  reverse(arr);
  println("P5",millis() - start);
  
  start = millis();
  Ru.reverse_copy(arr); 
  println("Rope with copy",millis() - start);
  
  start = millis();
  Ru.reverse(arr); 
  println("Rope",millis() - start);
  
  
  start = millis();
  reverse(arr);
  println("P5",millis() - start);
  
  start = millis();
  reverse(arr);
  println("P5",millis() - start);
  
  exit();
}

class Machin {
  Machin(){}
}
```

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -8252,15 +8252,49 @@ public class PApplet implements PConstants {
     return outgoing;
   }
 
-  static public Object reverse(Object list) {
-    Class<?> type = list.getClass().getComponentType();
-    int length = Array.getLength(list);
-    Object outgoing = Array.newInstance(type, length);
-    for (int i = 0; i < length; i++) {
-      Array.set(outgoing, i, Array.get(list, (length - 1) - i));
+  /*
+   * This function Object reverse(Object list)ccan be replace by <T> T [] reverse(T [] list) is more faster for every case
+   * 
+   */
+  // static public Object reverse(Object list) {
+  //   Class<?> type = list.getClass().getComponentType();
+  //   int length = Array.getLength(list);
+  //   Object outgoing = Array.newInstance(type, length);
+  //   for (int i = 0; i < length; i++) {
+  //     Array.set(outgoing, i, Array.get(list, (length - 1) - i));
+  //   }
+  //   return outgoing;
+  // }
+
+  static public <T> T [] reverse(T [] list) {
+    int len = list.length;
+    T[] outgoing = (T[]) Array.newInstance(list.getClass().getComponentType(), len);
+    for (int i = 0; i < outgoing.length; i++) {
+      outgoing[i] = list[len - i - 1];
     }
     return outgoing;
   }
+
+  static public long [] reverse(long [] list) {
+    int len = list.length;
+    long[] outgoing = new long[list.length];
+    for (int i = 0; i < outgoing.length; i++) {
+      outgoing[i] = list[len - i - 1];
+    }
+    return outgoing;
+  }
+
+  static public double [] reverse(double [] list) {
+    int len = list.length;
+    double[] outgoing = new double[list.length];
+    for (int i = 0; i < outgoing.length; i++) {
+      outgoing[i] = list[len - i - 1];
+    }
+    return outgoing;
+  }
+
+
+
 
 
 


benfry:
Thanks for looking into it—I know how frustrating it can be to spend this much time and not have it come up with much!

Unfortunately I don't think we can merge this. We're down to adding `double` and `long`, but we don't have those anywhere in `PApplet`. The array methods like `reverse` were in `PApplet` before we had the `XxxxList` classes, which is a better place for adding these features. (And indeed, `reverse` is already supported in `DoubleList` and `LongList` with this optimization already in place.)

And for the change from `Object` to `<T>`, that will break any existing Processing code (such as a library) that is using these methods until they have been recompiled. So unless there's an extremely strong reason to change (i.e. it's a heavily used method that has a bug), I have to avoid changing them. 

Sorry to not have better news but I hope you can understand the constraints we're juggling here.

knupel:
@benfry No problem, I understand well ; and it's not a time lost for me because I progress in a sametime :)  I try to back soon with another somethig maybe more utils and can be merge!

benfry:
Thanks, appreciate your understanding and the help as well!

PR #772, opened 2023-08-20
Full line edt

sampottinger:
We got pretty close on #752. However, the EDT often will place errors in unexpected places and with a width of 1 character (gives us a point where there was a problem but not the correct number of characters involved). This does not impact ANTLR issues.

I believe that, previous to the efforts to improve error highlighting, earlier versions of Processing 4 were highlighting the whole line but stuff related to #752 and #715 tried to make this more precise. This PR restores that earlier behavior and formalizes the interface. See attached.

Note that this does not impact upsteram errors in the build pipeline, only things that get through to the EDT. It also does not change behavior for errors reported at compile or runtime using javac directly.

before
after
after_vscode

### DIFF
--- a/app/src/processing/app/Problem.java
+++ b/app/src/processing/app/Problem.java
@@ -78,7 +78,7 @@ public interface Problem {
    * Get the exact character on which this problem ends in code line relative.
    * 
    * @return Number of characters past the start of the line if known where the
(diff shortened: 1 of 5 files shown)


benfry:
Looks good; thanks!

PR #786, opened 2023-09-23
Fix hash of keyEvent being added to pressedKeys.

### LINKED ISSUE #779: `keyPressed` is `true` with no keys pressed after specific sequence of `SHIFT` and any letter key
## Description

`keyPressed` is `true` with no keys actively pressed after doing a specific sequence of key presses with `SHIFT` and a letter key. It also occurs with `CTRL` and a letter key. This is more easily reproduced with P2D, but it does occur with the default renderer as well.

## Expected Behavior

`keyPressed` should be `false` when no keys are actively pressed.

## Steps to Reproduce

1. Press and hold a letter key
2. While still holding letter key, press and hold `SHIFT`
3. While still holding `SHIFT`, release the letter key
4. Release `SHIFT`

Pressing `CTRL` instead of `SHIFT` will also produce the behavior. It appears the only way to set `keyPressed` back to `false` is to press/release the specific letter key again. So if you did `SHIFT`+`a`, press and release `a` to set `keyPressed` back to the expected behavior. You can try to reproduce the behavior with this sketch:

```
color black = color(36,39,45);
color green = color(141,187,110);
void setup(){
  size(1600,900,P2D); 
}

void draw(){
  if (keyPressed){
    background(green);
  } else{
    background(black);
  }
}
```

## Your Environment

 

* Processing version: 4.3
* Operating System and OS version: Windows 10 Home 22H2 19045.3324

harry1064:
When we press a letter, lets say `d` is key is char value `d` and hash calculated using keyCodea and key is added to pressed keys. But when without release `d`, if we press `SHIFT` key, hash of shift key also being added to pressedKeys. Now, when we release `d`, it is actually now `D` hence key value change and when we try to remove new hash, it doesnot exist in pressedKeys and the old hash of `d` still stays in pressedKeys which was the cause of keyPressed returning true.

fixes #779

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -2627,16 +2627,15 @@ public class PApplet implements PConstants {
     keyEvent = event;
     key = event.getKey();
     keyCode = event.getKeyCode();
-
+    Long hash = ((long) keyCode << Character.SIZE);
     switch (event.getAction()) {
       case KeyEvent.PRESS -> {
-        Long hash = ((long) keyCode << Character.SIZE) | key;
         if (!pressedKeys.contains(hash)) pressedKeys.add(hash);
         keyPressed = true;
         keyPressed(keyEvent);
       }
       case KeyEvent.RELEASE -> {
-        pressedKeys.remove(((long) keyCode << Character.SIZE) | key);
+        pressedKeys.remove(hash);
         keyPressed = !pressedKeys.isEmpty();
         keyReleased(keyEvent);
       }


harry1064:
Hi @benfry 
Hope you are doing well. I saw in your [commit](https://github.com/processing/processing4/commit/1849ab76bcd1d10bee397c99775efd3415f4fa4b#diff-6abc5b99da96e317f0754499132ed82a1980faddf44317647b2a5e35d6c817d3R17-R19)  that you have included this pull request in `todo.txt`.
Does it mean you're going to have a look at it when you have time?

benfry:
Yes, went to pull it in but realized that I needed to do additional testing… This is a really messy issue because of the way that different key combinations are posted as events (and there are differences to this between Java2D, OpenGL, and JavaFX as well) so it just needs more thorough work to make sure this is doing what we need. I suspect that this change doesn't fix the problem broadly enough, and want to make sure we have it right before altering something so fundamental.

harry1064:
Yes, that make sense.. I was also curious why `| key` was done in first place.  Testing it more thoroughly  make sense.

benfry:
Yeah, it's because with any keystroke there may be a combination of `key` and `keyCode`, and those vary as different modifiers are applied. It's especially important in this scenario where we're trying to just detect what keys and modifiers are doing.

PR #826, opened 2024-03-13
Fix #825 - Showing wrong errors' line numbers for .java files

### LINKED ISSUE #825: Wrong line numbers in error messages when using .java files
## Description

The Processing IDE reports wrong source code location for a programming error in .java-files

The Processing IDE should change such that errors are reported correctly. This severely degrades the overall usability of the Processing IDE. I degrades programmer's productivity.

## Expected Behavior

When I include .java files in my sketch and I make programming errors, I would expect that the Processing IDE shows the errors at the right source code location.

## Current Behavior

Instead it reports the error in the same file but with the wrong line number.

## Steps to Reproduce

My sketch contains two files:
- `BrokenErrorMessage.pde`, whose content is irrelevant, and 
- `SomeJavaCodeWithErrors.java`:
  ```
  public class SomeJavaCodeWithErrors // line 1
  {                                   // line 2
    public static int testMe()        // line 3
    {                                 // line 4
      // provoking an error!          // line 5
      return asdflkj;                 // line 6, error, but it is reported be on line 3
    } 
  }
  ```
When I hit the "play" button in the Processing IDE, I, of course, do get an error:
`The variable "asdflkj" does not exist`, but the source code location refers to line `3`.

However, I would have expected the line number of the error _not_ to be `6` and instead be `3`.

## Your Environment

 

* Processing version: 4.3
* Operating System and OS version: Windows 10 Enterprise 22H2
* Other information:

In the Processing forum, we have already  discussed  the bug here https://discourse.processing.org/t/wrong-line-numbers-in-error-messages-when-using-java-files/43983 and concluded to raise an issue on github.

e1dem:
Fix for bug [#825](https://github.com/benfry/processing4/issues/825).

Using PreprocSketch.mapJavaToSketch result for processing .java files is incorrect, as a sketch's combined source code file does not include code from .java tabs.

As far as I can see, there is a similar issue with showing variables usage, so I would suggest this solution as a temporary one.


hkiel:
@benfry Is this PR going to be incorporated at any time soon?

